PR #23, opened 2014-11-06
add some fixes

nhanitvn:
1. to make the R package build-able
2. add db_list_tables and db_has_table for DDFManager
To-do: tbl does not work complaining that there is no db_list_tables for DDFManager class though that function works when it is called in a standalone way.


ctn:
Is this PR to be updated, or closed?

mbbui:
Too old to merge.

ctn:
@nhanitvn sup?

nhanitvn:
Let's close this. the dplyr code is not mature yet.

PR #69, opened 2015-03-18
Copy factor

Huandao0812:
put copyFactor implementation to core
copy factor information after transform
check for column before setting factor


core/src/main/java/io/ddf/content/AMetaDataHandler.java
```diff
@@ -59,9 +59,19 @@ public long getNumRows() throws DDFException {
...
-
-  };
+  /**
+   * Transfer factor information from ddf to this DDF
+   * @param ddf
+   * @throws DDFException
+   */
+  public void copyFactor(DDF ddf)  throws DDFException {
+    for (Schema.Column col : ddf.getSchema().getColumns()) {
+      if (this.getDDF().getColumn(col.getName()) != null && col.getColumnClass() == Schema.ColumnClass.FACTOR) {
```
mbbui: Good that you check null here. I will also return an error message so user know what is going on.

PR #81, opened 2015-04-15
Use Rsession instead of RConnection/RServe

dabaitu:
Motivation for this change:
Rsession is a wrapper over REngine/RServe that will start a RServe if there's not already one running. Currently we have to use an external cron job to keep a RServe running so DDF can pass tasks over for R to process, which is highly unreliable.
Using Rsession will rid of the need for external cron job and is much more reliable.

Please also see the JIRA ticket(https://adatao.atlassian.net/browse/PA-268)

@nhanitvn @mbbui please check this out lightly as I proceed, call me out on bad approach as early as possible. Thanks!

### DIFF
--- a/project/RootBuild.scala
+++ b/project/RootBuild.scala
@@ -191,7 +191,6 @@ object RootBuild extends Build {
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.slf4j" % "slf4j-log4j12" % slf4jVersion,
       "commons-configuration" % "commons-configuration" % "1.6",
-      "com.google.api-client" % "api-client" % "1.18.0-rc",
       "com.google.guava" % "guava" % "14.0.1",
       "com.google.code.gson"% "gson" % "2.2.2",
       //"org.scalatest" % "scalatest_2.10" % "2.1.0" % "test",
--- a/project/RootBuild.scala
+++ b/project/RootBuild.scala
@@ -191,6 +191,7 @@ object RootBuild extends Build {
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.slf4j" % "slf4j-log4j12" % slf4jVersion,
       "commons-configuration" % "commons-configuration" % "1.6",
+      "com.google.api-client" % "api-client" % "1.18.0-rc",
       "com.google.guava" % "guava" % "14.0.1",
       "com.google.code.gson"% "gson" % "2.2.2",
       //"org.scalatest" % "scalatest_2.10" % "2.1.0" % "test",
--- a/spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
+++ b/spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
@@ -29,8 +29,8 @@ class TransformationHandler(mDDF: DDF) extends CoreTransformationHandler(mDDF) {
     // Prepare data as REXP objects
     val dfrdd = mDDF.getRepresentationHandler.get(classOf[RDD[_]], classOf[REXP]).asInstanceOf[RDD[REXP]]
 
-	// RSession to use
-	val rsess = Rsession.newInstanceTry(mLog);
+    // RSession to use
+    val rsess = Rsession.newInstanceTry(mLog);
 
     // 1. map!
     val rMapped = dfrdd.map {
@@ -101,7 +101,7 @@ class TransformationHandler(mDDF: DDF) extends CoreTransformationHandler(mDDF) {
         try {
           // send the df.partition to R process environment
           val dfvarname = "df.partition"
-		  rsess.set(dfvarname, partdf)
+          rsess.set(dfvarname, partdf)
 
           val expr = String.format("%s <- transform(%s, %s)", dfvarname, dfvarname, transformExpression)
 
@@ -112,7 +112,7 @@ class TransformationHandler(mDDF: DDF) extends CoreTransformationHandler(mDDF) {
 
           // transfer data to JVM
           val partdfres = rsess.eval(dfvarname)
-		  rsess.rm(dfvarname)
+          rsess.rm(dfvarname)
 
           partdfres
         } catch {
--- a/spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
+++ b/spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
@@ -22,15 +22,17 @@ import io.ddf.etl.{TransformationHandler ⇒ CoreTransformationHandler}
 import io.ddf.exception.DDFException
 import io.spark.ddf.SparkDDF
 
+import java.io.PrintStream
+import com.google.api.client.util.LoggingOutputStream
+
 class TransformationHandler(mDDF: DDF) extends CoreTransformationHandler(mDDF) {
+  // Rsession to use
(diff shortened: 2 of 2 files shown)


khangich:
ok to test

project/RootBuild.scala
```diff
@@ -191,6 +191,7 @@ object RootBuild extends Build {
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.slf4j" % "slf4j-log4j12" % slf4jVersion,
       "commons-configuration" % "commons-configuration" % "1.6",
+      "com.google.api-client" % "api-client" % "1.18.0-rc",
```
dabaitu: @nhanitvn @Huandao0812 - Hi guys, what's the correct way to add this dependency?(com.google.api-client)

I get cannot resolve dependency error when I try to compile

spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
```diff
@@ -22,7 +22,12 @@ import io.ddf.etl.{TransformationHandler ⇒ CoreTransformationHandler}
 import io.ddf.exception.DDFException
 import io.spark.ddf.SparkDDF
 
+import java.io.PrintStream
+import com.google.api.client.util.LoggingOutputStream
+
 class TransformationHandler(mDDF: DDF) extends CoreTransformationHandler(mDDF) {
+  // Rsession to use
+  val rsess = Rsession.newInstanceTry(new PrintStream(new LoggingOutputStream(mLog, false), true), null)
```
dabaitu: @nhanitvn @Huandao0812 this is where I use LoggingOutputStream that comes from com.google.api-client. I have to redirect Rsession's log to go to mLog.

spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
```diff
@@ -96,26 +101,20 @@ class TransformationHandler(mDDF: DDF) extends CoreTransformationHandler(mDDF) {
     val rMapped = dfrdd.map {
```
dabaitu: After evaluating R exp with Rsession, rMapped became another type of variable:

[error]  found   : org.apache.spark.rdd.RDD[Nothing]
[error]  required: org.apache.spark.rdd.RDD[org.rosuda.REngine.REXP]

The error point to line 129 of this file(I can't comment on that line)

project/RootBuild.scala
ctn: http://mvnrepository.com/artifact/com.google.api-client/google-api-client/1.18.0-rc

spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
dabaitu: Resolved by
val partdfres: REXP = rsess.eval(dfvarname)

@Huandao0812 can you help me understand why I have to enforce the type? According the the function I shouldn't need to: https://github.com/yannrichet/rsession/blob/e728223b7bc38b0d737f978772ef0929d0b759f4/Rsession/src/org/math/R/Rsession.java#L1170

spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
dabaitu: temporarily bypassing this issue by dumping Rsession output to console: Rsession.newInstanceTry(System.out, null)

Experts: please help

project/RootBuild.scala
dabaitu: @ctn Yup. thank you! I actually had the wrong artifact name, should be google-api-client

ctn: But I showed you how to fish, right? :)

dabaitu: @ctn yes sir, can't wait to start whaling

spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
ctn: @dabaitu what likely explains this is as follows: one or more of the `rsess.eval()` can throw/threw an exception, which caused a `Nothing` to be returned to the Scala side, which then causes the entire `RDD` to be `RDD[Nothing]`. By expressly specifying `REXP` you ensure that the `RDD` is `RDD[REXP]`.

spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
aht: @dabaitu this generally looks good! It's seems RSession sensibly re-use most of the Rserve API e.g. REXP so the changes are not too bad.

What is really important is to understand the inter-process overall data flow to execute R code on top of DDF. See this doc: https://docs.google.com/a/adatao.com/document/d/1BOl-FU-jKijlMGUaYy4gC0SIo0KAYwr3-o_G3mdd_uo/edit

In this case, each `rsess` should be created inside the dfrdd.map so that each data partition gets a new R instance and the map tasks can proceed in parallel. (BTW, It's a map but it's really a mapPartition behind the scene, `partdf` is a whole partition represented as a REXP).

Re logging, this is a good question. When we start Rserve manually we redirect it to a log file on each slave nodes. Again, note that when you create the Rsession you are sitting at a slave node and presented a data partition to work on. 

@mbbui to suggest something sensible re logging. Redirect to stdout is nice if YARN is able to capture them from the slave executors.

dabaitu:
@aht @nhanitvn @mbbui 
I tested briefly with R, and seems the Rsession logs would get captured by PA log in /tmp/pa.out. From the log it seems things are working.
I had to export R_HOME=/usr, just so Rsession knows where the R executable is. I think this belongs with DDF but where do you guys think a config like this should go inside DDF repo? Currently I have it in pa-env.sh

spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
dabaitu: @ctn you are right. The eval call eventually call silentlyEval which in some cases returns null, but it's a null of type REXP. I guess Scala just auto-cast whatever null into it's Nothing type?

spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
dabaitu: @aht I tried to test with some R transforms, and it threw an error saying "task is not serializable" on the line where I construct Rsession inside the map(as in map-reduce). Do you know how we were able to do that with Rconnection? Mike suspected it was because of the System.out so I tried out another constructor for Rsession(val rsess = new Rsession(null, true)) and still getting the same error.

ctn:
So I'll close #81 until you're ready to reopen it?

khangich:
ok to test

dabaitu:
@ctn still working on this a bit. I was testing today and some of the transforms failed. Getting help from Mike and AHT.

spark/src/main/scala/io/spark/ddf/etl/TransformationHandler.scala
dabaitu: @aht resolved. I had the Rsession declaration in a bad place(like in the middle of the crack).

PR #82, opened 2015-04-16
Mutable ddf

Huandao0812:
Only TransformUDF is mutable


Huandao0812:
ok to test

Huandao0812:
ok to test

PR #84, opened 2015-04-21
add UUID to DDF

Huandao0812:


core/src/main/java/io/ddf/DDF.java
```diff
@@ -166,7 +171,9 @@ protected void initialize(DDFManager manager, Object data, Class<?>[] typeSpecs,
 
     this.setName(name);
 
+    this.uuid = UUID.randomUUID().toString();
```
darma-m: We should switch to time-based UUID (v1) post-GA.

core/src/main/java/io/ddf/DDFManager.java
```diff
@@ -77,36 +77,42 @@
...
-  // ephemeral mapping between ddf alias and DDF
-  protected HashMap<String, DDF> mDDFsByName = new HashMap<String, DDF>();
-
   protected Map<String, IModel> mModels = new HashMap<String, IModel>();
 
-  // lookup name (aliasName, name)
-  protected Map<String, String> mAliasesToNames = new HashMap<String, String>();
-
 
-  public String addDDF(DDF data) {
```
darma-m: Let's use "ddf" parameter name, instead of "data" - too overloaded.

core/src/main/java/io/ddf/DDFManager.java
```diff
@@ -77,36 +77,42 @@
...
-    mDDFs.put(data.getUri(), data);
-    return data.getUri();
+  public void addDDF(DDF ddf) throws DDFException {
+    mDDFs.put(ddf.getUUID(), ddf);
   }
 
-  public DDF getDDF(String ddfName) {
-    DDF data = mDDFs.get(ddfName);
+  public DDF getDDF(String uuid) {
+    DDF data = mDDFs.get(uuid);
```
darma-m: Let's use "ddf" variable name, instead of "data" - too overloaded.

PR #88, opened 2015-04-24
Improve histogram

namma:
ddf.getVectorHistogram is now using implementation getVectorHistogramImp inside file BinningHandler.scala

### DIFF
--- a/spark/src/main/scala/io/spark/ddf/analytics/BinningHandler.scala
+++ b/spark/src/main/scala/io/spark/ddf/analytics/BinningHandler.scala
@@ -1,17 +1,38 @@
 package io.spark.ddf.analytics
 
 import io.ddf.DDF
+import org.apache.spark.rdd.{DoubleRDDFunctions, RDD}
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.catalyst.expressions._
 import scala.collection.JavaConversions._
 import scala.collection.JavaConverters._
-import io.ddf.analytics.ABinningHandler
+import io.ddf.analytics.{AStatisticsSupporter, ABinningHandler, IHandleBinning}
 import io.ddf.analytics.ABinningHandler._
-import io.ddf.analytics.IHandleBinning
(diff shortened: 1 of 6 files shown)


khangich:
ok to test

namma:
it passed many unit tests, including DDF with missing values and/or invalid data. Please merge.

spark/src/main/scala/io/spark/ddf/analytics/BinningHandler.scala
```diff
@@ -1,18 +1,45 @@
...
 
 class BinningHandler(mDDF: DDF) extends ABinningHandler(mDDF) with IHandleBinning {
 
+
+  def parseDouble(r: Row) = try {r.get(0).toString.toDouble } catch { case _ => None }
+
+  override def getVectorHistogramImpl(columnName: String, numBins: Int): java.util.List[AStatisticsSupporter.HistogramBin] = {
+    val projectedDDF: DDF = mDDF.VIEWS.project(columnName)
+    val rdd: RDD[Row] = projectedDDF.getRepresentationHandler.get(classOf[RDD[_]], classOf[Row]).asInstanceOf[RDD[Row]]
+    val rdd1 = rdd.map(r => {try {r.get(0).toString.toDouble } catch { case _ => None }})
```
Huandao0812: bad way to parse double

namma: Can you elaborate? This one is to eliminate missing data or invalid value.
What is your suggestion?

mbbui: Huan, you should tell why it is bad. If Nam already knows why it is bad he would not do this already.

namma: I talked with Huan. He said try-catch significantly decreases the performance. I will check and update the code accordingly when I have cycle. At the moment, I believe it is not a stopper issue though.

PR #95, opened 2015-05-01
Properly handle when column with same name is added

dabaitu:
@mbbui @nhanitvn @Huandao0812 please review

### DIFF
--- a/core/src/main/java/io/ddf/etl/TransformationHandler.java
+++ b/core/src/main/java/io/ddf/etl/TransformationHandler.java
@@ -90,7 +90,6 @@ public class TransformationHandler extends ADDFFunctionalGroupHandler implements
     String sqlCmd = String.format("SELECT %s FROM %s",
         RToSqlUdf(RExp, columns, this.getDDF().getSchema().getColumns()),
         this.getDDF().getTableName());
-    System.out.println("Performing: " + sqlCmd);
 
     DDF newddf = this.getManager().sql2ddf(sqlCmd);
 
--- a/core/src/main/java/io/ddf/etl/TransformationHandler.java
+++ b/core/src/main/java/io/ddf/etl/TransformationHandler.java
@@ -10,6 +10,7 @@ import io.ddf.content.Schema.ColumnClass;
 import io.ddf.exception.DDFException;
 import io.ddf.misc.ADDFFunctionalGroupHandler;
 import java.util.List;
+import java.util.ArrayList;
 
 public class TransformationHandler extends ADDFFunctionalGroupHandler implements IHandleTransformations {
 
@@ -90,6 +91,10 @@ public class TransformationHandler extends ADDFFunctionalGroupHandler implements
     if (columns != null) {
       columnList = Joiner.on(",").skipNulls().join(columns);
     } else {
+      columns = new ArrayList<String>();
+      for (Column c : this.getDDF().getSchema().getColumns()) {
+        columns.add(c.getName());
+      }
       columnList = "*";
     }
     String sqlCmd = String.format("SELECT %s, %s FROM %s", columnList, RToSqlUdf(RExp, columns), this.getDDF().getTableName());
@@ -115,6 +120,7 @@ public class TransformationHandler extends ADDFFunctionalGroupHandler implements
    */
 
   public static String RToSqlUdf(String RExp, List<String> existingColumns) {
+    System.out.println("Existing columns: " + Joiner.on(",").skipNulls().join(existingColumns));
     List<String> udfs = Lists.newArrayList();
     for (String str : RExp.split(",(?![^()]*+\\))")) {
       String[] udf = str.split("[=~](?![^()]*+\\))");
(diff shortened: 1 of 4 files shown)


spark/src/test/java/io/spark/ddf/analytics/BinningHandlerTest.java
```diff
@@ -35,7 +35,7 @@ public void testBinning() throws DDFException {
     Assert.assertTrue(ddf1.getSchemaHandler().getColumn("month").getColumnClass() == ColumnClass.FACTOR);
     // {'[2,4]'=1, '(4,6]'=2, '(6,8]'=3}
     System.out.println(">>>>> get2,4) == " + ddf1.getSchemaHandler().getColumn("month").getOptionalFactor().getLevelMap().get("[2,4]"));
-	Assert.assertTrue(ddf1.getSchemaHandler().getColumn("month").getOptionalFactor().getLevelMap().get("[2,4]") == 2);
+	Assert.assertTrue(ddf1.getSchemaHandler().getColumn("month").getOptionalFactor().getLevelMap().get("[2,4]") == 1);
```
dabaitu: I need help understand what binning does. I changed this value without understanding what's happening so I can have a PR and ask a question

spark/src/test/java/io/spark/ddf/etl/TransformationHandlerTest.java
```diff
@@ -88,7 +88,7 @@ public void testReservedFactor() throws DDFException {
 
     Assert.assertEquals(31, ddf.getNumRows());
     Assert.assertEquals(9, ddf.getNumColumns());
-    Assert.assertEquals("test123", ddf.getColumnName(8));
+    Assert.assertEquals("test123", ddf.getColumnName(0));
     Assert.assertEquals(9, ddf.VIEWS.head(1).get(0).split("\\t").length);
```
dabaitu: new columns are now added to the front

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -85,14 +87,11 @@ public DDF transformMapReduceNative(String mapFuncDef, String reduceFuncDef, boo
   }
 
   public DDF transformUDF(String RExp, List<String> columns) throws DDFException {
+    String sqlCmd = String.format("SELECT %s FROM %s",
+        RToSqlUdf(RExp, columns, this.getDDF().getSchema().getColumns()),
+        this.getDDF().getTableName());
+    System.out.println("Performing: " + sqlCmd);
```
mbbui: use log pls

dabaitu:
ok to test

core/src/main/java/io/ddf/etl/TransformationHandler.java
dabaitu: @mbbui I should just remove this

spark/src/test/java/io/spark/ddf/etl/TransformationHandlerTest.java
nhanitvn: I see that new columns in transform using Hive are added to the front, while the new ones in transform using R are added to the back. We should be consistent. To me, adding them to the back is more preferred.

spark/src/test/java/io/spark/ddf/analytics/BinningHandlerTest.java
nhanitvn: Binning discretizes a continuous column into a categorical one. There are three ways of doing that: 1. create bins with equal size, 2. create bins with equal frequency, 3. create bins with custom size.

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -114,26 +112,45 @@ public DDF transformUDF(String RExp) throws DDFException {
    * @return "(arrtime - crsarrtime) as foobar, (distance / airtime) as speed
    */
 
-  public static String RToSqlUdf(String RExp, List<String> existingColumns) {
+  public static String RToSqlUdf(String RExp, List<String> selectedColumns, List<Column> existingColumns) {
```
nhanitvn: What if we have expressions with functions existing in R but not in Hive? Don't we support native R expression using Rserve anymore?

spark/src/test/java/io/spark/ddf/etl/TransformationHandlerTest.java
dabaitu: @nhanitvn , yes I discussed with Bao actually. I didn't think of the R transform part. The reason I left the columns as they are is because I don't want to introduce more latency(tiny ones add up). The logic right now is a bit convoluted because the API and the behavior we want to support is complicated. We can add/select/delete/redefine columns in the RExp, and we can also select columns in another parameter. Let me re-think and see if I can improve this.

core/src/main/java/io/ddf/etl/TransformationHandler.java
dabaitu: @nhanitvn We do support nativeR, but the user will have to specify "R". If "R" is specified, the expression is directly passed to transformNativeRserve. Again, I think here exist a mismatch between the API and what can be passed to the API. In this case, even though we specify "Hive" the expression we would give it is R, which doesn't really make sense to me.

PR #96, opened 2015-05-04
Keep column in order

dabaitu:
Besides move new columns added through transformHive to the end of the schema. Some more checks are added to defend against inconsistencies.

transformUDF takes in two parameters: RExp, and a list of selected columns

new columns can be added in both the RExp and the list, so we should defend against duplicate/conflicting columns in RExp or in the list or across them(throw exception). We should really get rid of the second parameter; it doesn't make sense to me and it require a lot more effort to make sure nothing weird happens.

Right now there's still one minor case left:
1. ddf.Transfrom.transformUDF("arrtime, speed=distance/airtime")
2. ddf.Transfrom.transformUDF("speed=distance/airtime")
These two will produce same output, because any existing columns selected in the RExp part(meaning just column name and no '=' followed by a new definition) will be effectively ignored, since RExp is a place to specify new definition for an existing column or a new column. A filter is done only if the a list is provided.

The above issue has always been there. Before, we could also duplicate columns through the list or the RExp.

@nhanitvn @mbbui please review.

### DIFF
--- a/core/src/main/java/io/ddf/etl/TransformationHandler.java
+++ b/core/src/main/java/io/ddf/etl/TransformationHandler.java
@@ -12,6 +12,8 @@ import io.ddf.misc.ADDFFunctionalGroupHandler;
 import java.util.List;
 import java.util.Map;
 import java.util.HashMap;
+import java.util.Set;
+import java.util.HashSet;
 
 public class TransformationHandler extends ADDFFunctionalGroupHandler implements IHandleTransformations {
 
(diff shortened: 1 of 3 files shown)


dabaitu:
ok to test

Huandao0812:
ok to test

dabaitu:
the number of unit tests failing are increasing without any changes...anyone know what's going on?

dabaitu:
ok to test

dabaitu:
ok to test

dabaitu:
ok to test

dabaitu:
ok to test

PR #99, opened 2015-05-13
add function to merge 2 ddfs

Huandao0812:


khangich:
do we need unit test ?

khangich:
also what happen if I merge two different structure DDFs ?

mbbui:
good question from khang @Huandao0812

PR #100, opened 2015-05-13
Support struct type and array type in DDF to support json data

namma:


khangich:
ok to test

mbbui:
How if user wants to run this with our ML algorithms?

mbbui:
we need to review all other executors, to see how supporting struct will affect other operators. For operator that does not work, we should throw an error and let users know what to do with it.

namma:
This is first step to at least allow users to load data from a complex/raw data source like JSON into a DDF. ML algo can only work on valid dataset. The users need to do some ETL from the raw DDF to have valid input data for ML algos. If the input data is invalid (e.g. some column is struct), it throws error.

core/src/main/java/io/ddf/content/Schema.java
```diff
@@ -401,12 +401,17 @@ public void setData(Object[] data) {
...
-        Object.class), //
-    LOGICAL(Boolean.class), //
+    STRING(String.class),
+    INT(Integer.class),
+    LONG(Long.class),
+    FLOAT(Float.class),
+    DOUBLE(Double.class),
+    BIGINT(Long.class), //
+    TIMESTAMP(Date.class, java.sql.Date.class, Time.class, Timestamp.class),
+    STRUCT(Object.class),
```
nhanitvn: Supporting a datatype means we must support operations on that datatype. For example, how can users refer to the struct's members? how can they access to an array's element?

PR #101, opened 2015-05-14
add support for grep ignoring case

nhanitvn:


spark/src/test/java/io/spark/ddf/content/ViewHandlerTest.java
```diff
@@ -31,4 +34,53 @@ public void testRemoveColumns() throws DDFException {
...
+    Operator grep = new Operator();
+    grep.setName(OperationName.grep_ic);
+    Expression[] operands = new Expression[2];
+    StringVal val = new StringVal();
+    val.setValue("iad");
+    operands[0] = val;
+    operands[1] = col;
+    grep.setOperarands(operands);
+
+    DDF ddf2 = ddf.VIEWS.subset(columns, grep);
```
mbbui: Nhan, how is grep_ic expressed in terms of API, I expect that subset command should have a parameter indicating ignoreCase = True | False, isn't it?

nhanitvn: Yes, but current way of representing Operator does not support that, so I have to do client check for ignore.case param and emit grep or grep_ic Operator as in https://github.com/adatao/RClient/pull/336

nhanitvn: This is in PyClient https://github.com/adatao/PyClient/blob/7d1d9d999c5dcc0efb6a67d919df7019083cc5b0/pAnalytics/adatao/core/subset_expression.py#L306

PR #104, opened 2015-05-19
remove DDFInformation

Huandao0812:
listDDFs return an Array of DDF


core/src/main/java/io/ddf/DDFManager.java
```diff
@@ -89,6 +88,10 @@ public void removeDDF(DDF ddf) throws DDFException {
     mDDFs.remove(ddf.getUUID());
   }
 
+  public DDF[] listDDFs() {
```
mbbui: @Huandao0812 will this return DDF without name as well? 
Also have you talked with Nhan/Vu to update his API.
Best way is to create task on jira for them and related it to your ticket.

Huandao0812: I updated RClient api, and talked with @khangich to update Narratives code

PR #105, opened 2015-05-21
Flatten DDF with struct columns

namma:
- make struct-column DDF from JSON
- get flattened columns from spark dataframe
- support selected subset of columns to be flattened
- make schema with flattened columns
- make flattened DDF from struct DDF


khangich:
ok to test

core/src/main/java/io/ddf/DDF.java
```diff
@@ -336,7 +336,7 @@ public Date getCreatedTime() {
   public List<String> sql2txt(String sqlCommand, String errorMessage) throws DDFException {
     try {
       sqlCommand = sqlCommand.replace("@this", this.getTableName());
-      return this.getManager().sql2txt(String.format(sqlCommand, this.getTableName()));
+      return this.getSqlHandler().sql2txt(String.format(sqlCommand, this.getTableName()));
```
mbbui: @Huandao0812: DDF may support other engines, so .getManager() is the right one.

Huandao0812: getSqlHandler is engines agnostic too, and at a correct level of abstraction.

PR #110, opened 2015-05-26
Support fetchRows, head, etc. for DDF with complex type columns 

namma:
sql2txt implementation has been updated. Value of a non-primitive column is now represented by JSON.

https://adatao.atlassian.net/browse/PE-514

### DIFF
--- a/spark/src/main/java/io/spark/ddf/etl/SqlHandler.java
+++ b/spark/src/main/java/io/spark/ddf/etl/SqlHandler.java
@@ -123,7 +123,7 @@ public class SqlHandler extends ASqlHandler {
     for (Row row : arrRow) {
       lsString.add(row.mkString("\t"));
     }
-    return lsString
+    return lsString;
     */
 
     String[] strResult = SparkUtils.jsonForComplexType(rdd, "\t");
--- a/spark/src/main/java/io/spark/ddf/etl/SqlHandler.java
+++ b/spark/src/main/java/io/spark/ddf/etl/SqlHandler.java
@@ -12,17 +12,23 @@ import io.ddf.exception.DDFException;
 import io.spark.ddf.SparkDDF;
 import io.spark.ddf.SparkDDFManager;
 import io.spark.ddf.content.SchemaHandler;
+import io.spark.ddf.util.SparkUtils;
 import org.apache.spark.rdd.RDD;
 import org.apache.spark.sql.DataFrame;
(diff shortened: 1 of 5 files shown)


khangich:
ok to test

spark/src/main/java/io/spark/ddf/etl/SqlHandler.java
```diff
@@ -109,12 +115,19 @@ public DDF sql2ddf(String command, Schema schema, String dataSource, DataFormat
...
     List<String> lsString = new ArrayList<String>();
 
     for (Row row : arrRow) {
       lsString.add(row.mkString("\t"));
     }
-    return lsString;
+    return lsString
+    */
+
+    String[] strResult = SparkUtils.jsonForComplexType(rdd, "\t");
```
Huandao0812: use DataFrame.toJson to convert the DataFrame to RDD[String]
https://github.com/apache/spark/blob/branch-1.3/sql/core/src/main/scala/org/apache/spark/sql/DataFrame.scala#L1236

@namma

spark/src/main/scala/io/spark/ddf/util/SparkUtils.scala
```diff
@@ -119,6 +124,74 @@ object SparkUtils {
...
+   * @param df
+   * @return an Array of string showing the dataframe with complex column-object replaced by json string
+   */
+  def jsonForComplexType(df: DataFrame, sep: String): Array[String] = {
+    val schema = df.schema
+    val df1: RDD[String] = df.map(r => rowToJSON(schema, r, sep))
+    df1.collect()
+  }
+
+  private def rowToJSON(rowSchema: StructType, row: Row, separator: String): String = {
```
Huandao0812: don't need to do this, use https://github.com/apache/spark/blob/branch-1.3/sql/core/src/main/scala/org/apache/spark/sql/DataFrame.scala#L1236

spark/src/main/java/io/spark/ddf/etl/SqlHandler.java
namma: no, it is different. It is to convert entire row to JSON. We only convert each column object to JSON.

spark/src/main/scala/io/spark/ddf/util/SparkUtils.scala
namma: are you sure ;)

PR #113, opened 2015-05-28
fix writing ddf to text used in sql2txt and fetchRows, head, etc. 

namma:
Fix this issue: https://adatao.atlassian.net/browse/PE-577

**Cause of the issue**: with complex data type in DDF, it uses general json writer to print out the content of DDF which add quotes to every string. 

**Result of the fix**
- string value is no longer in quote
- column with null value is included as 'null' instead of being ignored

### DIFF
--- a/spark/src/main/scala/io/spark/ddf/util/SparkUtils.scala
+++ b/spark/src/main/scala/io/spark/ddf/util/SparkUtils.scala
@@ -182,11 +182,18 @@ object SparkUtils {
     var i = 0
     rowSchema.zip(row.toSeq).foreach {
       case (_, null) =>
+        if(i > 0)
+          gen.writeRaw(separator)
+        i = i+1
+        gen.writeNull()
       case (field, v) =>
         if(i > 0)
           gen.writeRaw(separator)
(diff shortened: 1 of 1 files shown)


khangich:
ok to test

nhanitvn:
Could you add a regression test?

namma:
for example?
I tested quite carefully, to be sure that the result is the same with what we had before complex type DDF.

nhanitvn:
Convert your manual tests into test cases. Those test cases are named
"regression", that is to ensure a bug is fixed.

On Fri, May 29, 2015 at 1:09 AM, Nam <notifications@github.com> wrote:

> for example?
> I tested quite carefully, to be sure that the result is the same with what
> we had before complex type DDF.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ddf-project/DDF/pull/113#issuecomment-106541119>.
>



-- 
If life gives you lemons, make lemonade.

mbbui:
I think, in this case, a test to make sure the string return does not have quote. 
I will merge it now, but @namma can you still have the test.

PR #115, opened 2015-05-31
DDFSPARK_JAR now can contain directories in addition to files

nhanitvn:
In case of Databricks Cloud, the uploaded jar file is renamed by DBC by adding an UUID prefix. Also, itdoes not exist on the master node at the time init scripts are run, we can only know the directory where it will be copied, not the exact full path. So it is necessary that we can just specify the directory and let the DDF app search for the jar at run-time.


nhanitvn:
ok to test

nhanitvn:
retest this please

nhanitvn:
retest this please

nhanitvn:
retest this please

PR #116, opened 2015-06-01
DDF.getUri return null when name is null

Huandao0812:


khangich:
ok to test

Huandao0812:
ok to test

Huandao0812:
retest this please

PR #130, opened 2015-06-13
support parquet file format, make DataFormat convertable from int

dabaitu:
Dependent on by pqtFormat pull request from panalytics

### DIFF
--- a/core/src/main/java/io/ddf/content/Schema.java
+++ b/core/src/main/java/io/ddf/content/Schema.java
@@ -540,15 +540,15 @@ public class Schema implements Serializable {
 
     public static DataFormat fromInt(int x) {
       switch(x) {
-        case 1:
+        case 0:
           return SQL;
-        case 2:
+        case 1:
           return CSV;
-        case 3:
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/ddf/content/Schema.java
```diff
@@ -536,7 +536,24 @@ public static ColumnClass get(ColumnType type) {
    *
    */
   public enum DataFormat {
-    SQL, CSV, TSV, JSON
+    UNDEF, SQL, CSV, TSV, JSON, PQT;
```
ctn: Better to use `UNDEF(-1), SQL(0), CSV(1), TSV(2), JSON(3), PQT(4)`. Also you want SQL(0) to maintain backward compat.

Huandao0812:
why is DataFormat belong to DDF Schema?

dabaitu:
@Huandao0812 , I'm not sure, it was like that before. I think it shouldn't even be part of DDF, should be in just PE.

PR #124, opened 2015-06-05
new implementation of getRandomSample (sample2ddf) to make it work with complex data type in DDF

namma:

PR #119, opened 2015-06-02
Standardizing output of sql2txt with all dependencies addressed

namma:
Output of sql2txt is now in general/standard format. If a column is array, the output is "[..]". If a column is struct, the output is a json string. 

There are 5 dependencies on the old sql2txt. They have been updated wrt the new ouput format.
- getVectorQuantiles
- getFiveNumSummary
- 2 internal functions inside BinningHandler.java
- getVectorHistogram (using HiveSQ)

In which, getVectorHistogram is totally rewritten to avoid using sql2txt. 

All DDF unit tests are passed.

PR #114, opened 2015-05-29
Move back to old sql2txt for backward compatability

namma:

PR #107, opened 2015-05-22
complete flattenDDF with bugs fixed for corner cases plus more comprehensive tests

namma:
- Correct column name to have valid DDF from JSON
- Fix Bugs
- Add more tests using sleep_data and Smurf event logs. 
- Refactor code

PR #131, opened 2015-06-14
implementation of new type representation and conversion for DDF

namma:
References for problem statement and design:
[1] Data Type Support Across Adatao Stacks, https://docs.google.com/document/d/1Iq8cWsoQ7uY7HUUN0Mzg8dX3asSOhIRmw0CWVxGYslw/edit#heading=h.ytm1c0ic3kje
[2] [Design] Type Support DDF / PE / BA, https://docs.google.com/spreadsheets/d/1Tr4YDAVx9mdRbtz_gLdeFs5rBZ0bF7bs5NM-mQtzhQY/edit#gid=0
Related:
[3] Complex type for JSON data, https://docs.google.com/document/d/1h2CvHzg_vifN8qSYTL_AS6VbI9ZD6iDOQvf8fjx_KDo/edit#heading=h.kp99bp86f19z

Main JIRA tickets: 
https://adatao.atlassian.net/browse/PE-572: to support new type representation and conversion
https://adatao.atlassian.net/browse/PE-514: to support non-primitive type in DDF/PA and clients.
Related tickets are linked to these 2 main tickets.

This is breaking change in DDF. However, all dependencies are already updated to support this change via the following PRs:
#124 new implementation of getRandomSample (sample2ddf) to make it work with complex data type in DDF
#119  Standardizing output of sql2txt with all dependencies addressed
#114 Move back to old sql2txt for backward compatability 
#113 fix writing ddf to text used in sql2txt and fetchRows, head, etc.
#110 Support fetchRows, head, etc. for DDF with complex type columns
#107 complete flattenDDF with bugs fixed for corner cases plus more comprehensive tests
#105 Flatten DDF with struct columns
#100 Support struct type and array type in DDF to support json data


khangich:
ok to test

ctn:
@trulite @shiti what do you think of this from the perspective of ddf-flink?

Shiti:
@ctn I have requested access to the google docs mentioned in the PR. Otherwise if someone can post the relevant sections here, it will be easier for us to understand this PR in that context.

core/src/main/java/io/ddf/analytics/AStatisticsSupporter.java
```diff
@@ -183,35 +178,6 @@ public double getVectorCovariance(String xColumnName, String yColumnName) throws
...
-        return bins;
-      }
-    }
-    return null;
-  }
-*/
-
-
   private double parseDouble(String s) {
     mLog.info(">>>> parseDouble: " + s);
```
ctn: Do we need/want this `mLog` anymore?

core/src/main/java/io/ddf/content/Schema.java
```diff
@@ -398,24 +398,32 @@ public void setData(Object[] data) {
 
 
   /**
-   * The R concept of a column "type", such as STRING, INT, LONG, etc.
+   * The R concept of a column "type", such as STRING, INT, BIGINT, etc.
    */
   public enum ColumnType {
```
mbbui: PS: each engine may use different type system to represent their data, while this enum only assume java type.
Solution: we should make this become Abstract Class. Then for each engine we will subclass with specific implementation of type conversion.

mbbui: @Shiti what do you think?

Shiti: I think that would be better approach since different engines support different types. It will keep the DDF API cleaner and easier for supporting other engines.

PR #133, opened 2015-06-17
add Datasource for DDF

Huandao0812:


khangich:
ok to test

core/src/main/java/io/ddf/datasource/DataFormat.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+public enum DataFormat {
+  UNDEF(-1), SQL(0), CSV(1), TSV(2), JSON(3), PARQUET(4);
+
+  private int value;
+
+  private DataFormat(int value) {
+    this.value = value;
+  }
```
dabaitu: We'd need this to be able to convert int to a DataFormat:

    public static DataFormat fromInt(int x) {
      switch(x) {
        case 0:
          return SQL;
        case 1:
          return CSV;
        case 2:
          return TSV;
        case 3:
          return JSON;
        case 4:
          return PQT;
        default:
          return UNDEF;
      }
    }

dabaitu:
I'll update my PR for parquet after this one is merged. This is good.

core/src/main/java/io/ddf/datasource/DataFormat.java
dabaitu: The above conversion has to exist somewhere, and I think it fits here. Also we should leave UNDEF for default case and not limit to -1, since what if some weird number got in and it's not 0-4 or -1.

dabaitu:
There's some change that's causing accessing HiveContext in BaseTest to hang. UTs are not passing.

PR #135, opened 2015-06-22
fix bug in MutabilityHandler …

Huandao0812:
where a ddf with name cannot be transformed more than 1 time
add test


khangich:
ok to test

PR #139, opened 2015-06-28
Type revision (cont.): complete more dependencies to the new type system

namma:


khangich:
ok to test

namma:
ok to test

PR #146, opened 2015-07-07
Update fillNA to not project on columns which are considered for filling missing values

nhanitvn:
Previously, specifying columns param will result in a DDF containing only specified columns. That is not a n expected behavior. The columns param should be used to specify columns to be considered for filling missing values.


lebinh:
retest this please

PR #151, opened 2015-07-20
Copy ddf

Huandao0812:
add function to copy ddf


khangich:
ok to test

ctn:
Cool. FYI keep in mind a near future when DDFs will be moved/copied across different engines on demand.

PR #155, opened 2015-07-28
Add a custom Spark UDF for converting datetime strings to the standard Hive one

nhanitvn:
which is yyyy-MM-dd HH:mm:ss using Joda library
The API signature is date_parse(String datetime_string, String format)


nhanitvn:
ok to test

PR #174, opened 2015-08-17
Escape tab characters in string values to avoid confusion with the tab delimiters in sql() API

nhanitvn:


lebinh:
retest this please

PR #199, opened 2015-09-23
use = not == in SQL expressions

ubolonton:


khangich:
confirmed, it works on redshift/spark

PR #201, opened 2015-09-26
Add copyFactor with columns to re-compute factor info

nhanitvn:
Update copyFactor logic to copy existing factor info for columns not to
recompute factor info
Add setFactorLevels to set factor levels and counts to a factor columns
Refactor ViewHandler's subset to separate subset logic from copyFactor


core/src/main/java/io/ddf/content/AMetaDataHandler.java
```diff
@@ -112,10 +126,19 @@ public void copyFactor(DDF ddf)  throws DDFException {
...
-                ().toString());
+            ().toString());
       }
       if (this.getDDF().getColumn(col.getName()) != null && col.getColumnClass() == Schema.ColumnClass.FACTOR) {
+        // Set corresponding column as factor
         this.getDDF().getSchemaHandler().setAsFactor(col.getName());
+
+        // if not in list of columns to re-compute factors
+        // then we just copy existing factor info to the new ones
+        if (!columns.contains(col.getName())) {
```
khangich: cool :+1:

PR #213, opened 2015-10-11
disallow duplicated column name

Huandao0812:


core/src/test/java/io/ddf/content/SchemaTest.java
```diff
@@ -47,4 +48,11 @@ public void testParseSchema() {
       assert(column.getType().toString() != null);
     }
   }
+
+  @Test(expected = DDFException.class)
```
nhanitvn: nice, :)

PR #218, opened 2015-10-17
add fix for namespace

PangZhi:
Fix for the namespace problem. The namespace should be read from the configuration file.

/cc @nhanitvn


nhanitvn:
@PangZhi Need to merge this to master too.

PR #219, opened 2015-10-19
fix array in sql parser

PangZhi:
This PR fix the array syntax in parser:
(1) select split(col, '\t')[0] from tablename
(2) select column[0] from tablename

Jiras:
https://adatao.atlassian.net/browse/PE-1016
https://adatao.atlassian.net/browse/PE-1075

/cc @Huandao0812 @nhanitvn @mbbui


hai-adatao:
Tested with:
- split(col, ' ')[0]
- split(split(col, ' ')[0], ' ')[0]

PR #220, opened 2015-10-25
Fix namespace

nhanitvn:
Related PR: https://github.com/ddf-project/DDF/pull/218


PangZhi:
Fix the issue that current namespace can't be configured.

PR #223, opened 2015-10-30
add model support in ddf

steveyan777:
@Huandao0812 @nhanitvn please review the code:
This is the temp model persistence code. When Add Model, it will add the model to HashMap cache (existing), as well as to hdfs with override if exists in hdfs already, the location is similar to the DDF persistence in hdfs, with post fix ".model". When read a model, it checks the HashMap cache first, if not in cache, it read from hdfs, and add to cache.

@mbbui the code is in the DDF repo, which will be open sourced, should it be in pAnalytics/DDF?

Thanks!

Steve

### DIFF
--- a/core/src/main/java/io/ddf/DDFManager.java
+++ b/core/src/main/java/io/ddf/DDFManager.java
@@ -17,6 +17,7 @@ package io.ddf;
 
 
 import com.google.common.base.Strings;
+import io.basic.ddf.content.ModelPersistenceHandler;
 import io.ddf.content.APersistenceHandler.PersistenceUri;
 import io.ddf.content.IHandlePersistence.IPersistible;
 import io.ddf.content.IHandleRepresentations;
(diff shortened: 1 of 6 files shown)


nhanitvn:
1. Why do we assume HDFS here?
2. We may refactor to move Persistence out of DDF, according to sameer suggestion, :)

core/src/main/java/io/ddf/DDFManager.java
```diff
@@ -215,12 +216,39 @@ public DDF getDDFByURI(String uri) throws DDFException {
...
+   * Add the model to cache, as well as serialize it to hdfs (if it is already in hdfs, overrite it)
+   * @param model
+   */
   public void addModel(IModel model) {
     mModels.put(model.getName(), model);
+
+    try {
+      ModelPersistenceHandler.persistModel(model, true);
+    } catch (DDFException e){
+
```
nhanitvn: Should do sth here, either logging or throw an Exception.

PR #240, opened 2015-12-08
[PE-1236] fixed bugs with null values

phvu:
Handle null values in `PyObj2ArrayObject` and `TransformationHandler`.


phvu:
@hai-adatao please help verify. Thanks.

PR #242, opened 2015-12-14
Fix/blank double quote

PangZhi:
This PR is to fix the problem that "" can't be used in current sql. 
The SQL Parser is upgraded.

/cc @Huandao0812 @nhanitvn @hai-adatao


hai-adatao:
UT passed @Huandao0812 do you have any comment, if not we can merge this
@PangZhi

PR #252, opened 2015-12-25
add getter for uuid & name in IGloballyAddressable

Huandao0812:


Huandao0812:
pls review @PangZhi @phvu @nhanitvn @mbbui

PR #253, opened 2015-12-29
update sql parser for function support

PangZhi:
This PR is to update SQL support for Function, and it fixes the following query:
SELECT id,if(loan_status like "%Late%" or loan_status like "%Charged%" or loan_status " +
            "like "%Default%" or loan_status like "%Grace%", 1, 0) as risk,risk_category FROM {1}

/cc @hai-adatao @Huandao0812 @nhanitvn

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -5,7 +5,7 @@
     <artifactId>ddf_core_2.10</artifactId>
     <packaging>jar</packaging>
     <description>ddf_core</description>
-    <version>1.4.7-SNAPSHOT</version>
+    <version>1.4.8-SNAPSHOT</version>
     <name>ddf_core</name>
     <organization>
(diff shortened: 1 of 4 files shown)


adatao-ddf:
Which Jira ticket does it fix? @jing

PangZhi:
I didn't include jira here because I think it's better not to include company jira info in open-source project.

adatao-ddf:
Ah, right. Thanks.

Huandao0812:
@PangZhi you shouldn't change the pom.xml in core and spark folder, it'll be generated from Rootbuild.scala

hai-adatao:
@Huandao0812 we didn't exclude pom.xml file in .gitignore, did we? The rootbuild is at 1.4.8 now so I suppose this change is ok

PR #254, opened 2015-12-30
fix null by quoting null string for spark

PangZhi:
This PR is a temporary fix for the following issue:
As SqlResult contains only string, so we can't distinguish null value and "null" string. This fix quotes the string value by ''. So null value will become null, "null" string will become 'null'.

The situation that the string value itself contains '' has not be considered.

/cc @hai-adatao @ubolonton @nhanitvn @Huandao0812

### DIFF
--- a/spark/src/main/scala/io/ddf/spark/util/SparkUtils.scala
+++ b/spark/src/main/scala/io/ddf/spark/util/SparkUtils.scala
@@ -160,7 +160,13 @@ object SparkUtils {
         if(v == null)
           gen.writeNull()
         else if(!complexTypes.contains(simpleTypeString))
-          gen.writeRaw(v.toString.replaceAll("\t", "\\\\t"))
+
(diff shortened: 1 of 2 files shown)


PangZhi:
retest this please

hai-adatao:
My opinion is that the bug is not that big to have a fix in DDF / PyClient and RClient, this fix is a bit unpredictable even if we adapt this fix in PyClient / RClient. Also, this looks more like a workaround to me rather than a fix. 

This is not a blocking bug and the fix is kinda a breaking change. I suggest we mark this bug as techdebt and fix it later, when we have a better serialization mechanism (which support strong typed rather than tsv), this will be fixed easily (or automatically).

hai-adatao:
Any updates on this? Will we go ahead with this fix?

hai-adatao:
Let's fix this later, when we support strong typed

PR #261, opened 2016-01-13
validate invalid column name

Huandao0812:


core/src/main/java/io/ddf/DDF.java
```diff
@@ -346,6 +346,12 @@ private void validateSchema(Schema schema) throws DDFException {
         if (columnSet.contains(column.getName())) {
```
nhanitvn: Should we make this validation a Schema's method so that we can create a unit test for it without the need to instantiate any DDF?

core/src/main/java/io/ddf/DDF.java
```diff
@@ -346,6 +346,12 @@ private void validateSchema(Schema schema) throws DDFException {
...
           throw new DDFException(String.format("Duplicated column name %s", column.getName()));
         } else {
+          Pattern p = Pattern.compile("^[a-zA-Z0-9_-]*$");
```
hai-adatao: Should we force the user to start with character and end with character / digit. E.g: these colnames are considered invalid:
_colname
-colname
colname_
colname-

BA doesn't allow such colnames (but one conflict thing is some auto-generated colname is _c0, _c1....)

@nhanitvn @PangZhi ?

PR #262, opened 2016-01-15
clean sql api, replace datasource param as a boolean

PangZhi:
This PR is to change the sql api.
The confusing datasource param is replaced with a boolean value queryOnDDF.

/cc @nhanitvn @hai-adatao @Huandao0812

### DIFF
--- a/core/src/main/java/io/ddf/DDFManager.java
+++ b/core/src/main/java/io/ddf/DDFManager.java
@@ -374,9 +374,9 @@ public abstract class DDFManager extends ALoggable implements IDDFManager, IHand
     return this.sql2ddf(command, null, null, null);
   }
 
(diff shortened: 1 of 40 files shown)


hai-adatao:
@lebinh can you help take a look?

lebinh:
@PangZhi this does clear up some confusion in our SQL api but I believe the different between (1) query on DDF and (2) query on the data source are big enough to warrant 2 different APIs instead of just a parameter.

PangZhi:
@lebinh I suggest keep this first. Because new apis need more changes and we can refactor it later. Currently the open source version and the product version are diverged.

lebinh:
@PangZhi agree then, we just need to keep that in mind when refactoring DDFManager.

PR #264, opened 2016-01-19
Fix/python shell

PangZhi:
This PR is to sync python shell and R shell with updated sql api of ddf.

/cc @hai-adatao @phvu @nhanitvn


phvu:
lgtm

hai-adatao:
lgtm

nhanitvn:
@PangZhi Could we set a default value (either True or False) for queryOnDDF param?

nhanitvn:
lgtm

PR #267, opened 2016-01-25
fix getRandomSample with replacement when fraction > 1 and size > original size 

nhanitvn:
Change-Id: I755d1f7fdcc0a51afd2b3a9f16edb264c4eed551

### DIFF
--- a/spark/src/test/java/io/ddf/spark/content/ViewHandlerTest.java
+++ b/spark/src/test/java/io/ddf/spark/content/ViewHandlerTest.java
@@ -82,4 +82,22 @@ public class ViewHandlerTest extends BaseTest{
     DDF ddf2 = ddf.VIEWS.subset(columns, grep);
     Assert.assertEquals(2, ddf2.getNumRows());
   }
+
+  @Test
+  public void testOversampling() throws DDFException {
+    createTableAirline();
+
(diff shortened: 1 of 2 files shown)


spark/src/test/java/io/ddf/spark/content/ViewHandlerTest.java
```diff
@@ -82,4 +82,85 @@ public void testSubsettingWithGrepIgnoreCase() throws DDFException {
     DDF ddf2 = ddf.VIEWS.subset(columns, grep);
     Assert.assertEquals(2, ddf2.getNumRows());
   }
+
+  @Test
+  public void testSampleWithSize() throws DDFException {
+    createTableAirline();
+
+    DDF ddf = manager.sql2ddf("select * from airline", "SparkSQL");
+    int sampleSize = 2;
```
hai-adatao: Can we use sampleSize bigger than ddf.nrow, and make sure that without replacement, the return sample has nrow == ddf.nrow, and with replacement, the sample has nrow == sampleSize

hai-adatao: Never mind, didn't notice about over sampling test below :D

PR #268, opened 2016-01-25
Flatten ArrayType column 

binhmop:
https://adatao.atlassian.net/browse/PE-1482 
@PangZhi @Huandao0812

### DIFF
--- a/core/src/main/java/io/ddf/etl/TransformationHandler.java
+++ b/core/src/main/java/io/ddf/etl/TransformationHandler.java
@@ -7,10 +7,10 @@ import io.ddf.DDF;
 import io.ddf.analytics.Summary;
 import io.ddf.content.Schema.Column;
 import io.ddf.content.Schema.ColumnClass;
+import io.ddf.content.Schema.ColumnType;
 import io.ddf.datasource.SQLDataSourceDescriptor;
 import io.ddf.exception.DDFException;
 import io.ddf.misc.ADDFFunctionalGroupHandler;
-
 import java.util.*;
 
 public class TransformationHandler extends ADDFFunctionalGroupHandler implements IHandleTransformations {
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -106,8 +106,37 @@ public DDF flattenDDF() {
...
+            "Unable to fetch the first value of the requested column").getRows().get(0);
+
+    int arrSize = Integer.parseInt(arrSizeStr);
+    for (int i = 0; i < arrSize; i++) {
+
+      newCols.append(String.format(" %s[%d] as %s_c%d,", colName, i, colName, i));
+    }
+
+    newCols.setLength(newCols.length() - 1);
+    String sqlCmd = String.format("SELECT *, %s FROM @this", newCols.toString());
```
PangZhi: @binhmop Should we replace the old column or just append new ones?
For example, my original columns are [col1, col_array], new ones should be [col1, col_array, col_array_c0, col_array_c1] or [col1, col_array_c0, col_array_c1] ?

binhmop: it's up to the user to decide, they can use removeColumn to delete if they want

hai-adatao:
@binhmop do we support array of array / array of struct? If yes, can we have some tests for those? Also, please add a test for flatten non-array type.

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -106,8 +106,37 @@ public DDF flattenDDF() {
...
+    DDF newddf = this.getDDF().sql2ddf(sqlCmd);
+
+    if (this.getDDF().isMutable()) {
```
hai-adatao: @Huandao0812 @binhmop 
We don't have mutable logic anymore, all DDF should be mutable by default. Should we remove this unnecessary logic?

spark/src/test/java/io/ddf/spark/etl/TransformationHandlerTest.java
```diff
@@ -81,6 +81,34 @@ public void testTransformScaleStandard() throws DDFException {
     Assert.assertEquals(8, newddf1.getSummary().length);
   }
 
+  @Test
+  public void testFlattenArrayTypeColumn() throws DDFException {
+
+    ddf.setMutable(false);
```
hai-adatao: If we keep the mutable logic and can't write tests for both mutable = false and mutable = true, then mutable = true should have higher priority, since that's the only thing users can see now

core/src/main/java/io/ddf/etl/TransformationHandler.java
binhmop: Mutability is always exposed at the DDF layer.

PR #270, opened 2016-01-29
lock a DDF for method that mutate it

Huandao0812:


core/src/main/java/io/ddf/content/SchemaHandler.java
```diff
@@ -94,12 +94,12 @@ public Schema generateSchema() throws DDFException {
   }
 
   @Override
-  public void setFactorLevelsForStringColumns(String[] xCols) throws DDFException {
+  public synchronized void setFactorLevelsForStringColumns(String[] xCols) throws DDFException {
```
steveyan777: @Huandao0812 @hai-adatao let's make sure to run some performance test, will the extra overhead acceptable?

Steve

PR #280, opened 2016-03-08
Refactor/ddf2

zkidkid:
Please DON'T MERGE. This PR is only for code review.

Change for Schema & IDataSource.
Basically all ddf will be created from an IDataSource
In case of engine doesn't support a concrete datasource -> throw unsupported datasource exception
else Engine will do prepare and make ddf on this datasource.

/cc @sameerp @mbbui @steveyan777 @Huandao0812 @nhanitvn @hai-adatao @lebinh

### DIFF
--- a/core/src/main/java/io/ddf/content/Schema.java
+++ b/core/src/main/java/io/ddf/content/Schema.java
@@ -1,655 +1,193 @@
-package io.ddf.content;
+package io.ddf2.datasource.schema;
 
+import io.ddf2.DDFException;
+import io.ddf2.DDF;
 
-import com.google.common.base.Function;
-import com.google.common.base.Joiner;
-import com.google.common.base.Strings;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
-import com.google.gson.annotations.Expose;
-import io.ddf.Factor;
-import io.ddf.exception.DDFException;
-import io.ddf.util.Utils;
+import org.apache.http.annotation.NotThreadSafe;
 
-import java.io.Serializable;
-import java.sql.Time;
-import java.sql.Timestamp;
+import java.math.BigDecimal;
+import java.sql.*;
 import java.util.*;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 
 /**
- * Table schema of a DDF including table name and column metadata
+ * Created by sangdn on 1/4/16.
  */
-@SuppressWarnings("serial")
-public class Schema implements Serializable {
-  @Expose
-  private String mTableName;
-  @Expose
-  private List<Column> mColumns = Collections
-      .synchronizedList(new ArrayList<Column>());
+@NotThreadSafe
+public  class Schema {
 
-  private DummyCoding dummyCoding;
+  protected List<String> colNames;
+  protected List<Column> columns;
 
-  /**
-   * Constructor that can take a list of columns in the following format:
-   * "<name> <type>, <name> <type>". For example,
-   * "id string, description string, units integer, unit_price float, total float"
-   * . This string will be parsed into a {@link List} of {@link Column}s.
-   *
-   * @param columns
-   */
-  @Deprecated
-  // Require tableName at all times, even null
-  public Schema(String columns) throws DDFException {
-    this.initialize(null, this.parseColumnList(columns));
+  public Schema() {
+    colNames = new ArrayList<>();
+    columns = new ArrayList<>();
   }
 
-  /**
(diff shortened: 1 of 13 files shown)


zkidkid:
@sameerp @PangZhi @mbbui @nhanitvn @Huandao0812 @lebinh @phvu could you help to review & comment on it.
more doc: https://docs.google.com/document/d/1iTAUREZnfua1xMEvjLpC4CO69zaXI-74QANRDqZnn9s/edit

core/src/main/java/io/ddf/content/Schema.java
```diff
@@ -1,655 +1,193 @@
...
+    columns = new ArrayList<>();
   }
 
-  public String getTableName() {
-    return mTableName;
+  public Schema(Schema schema) {
+    this();
+    this.colNames.addAll(schema.colNames);
+    columns.addAll(schema.getColumns());
+    assert colNames.size() == columns.size();
```
sameerp: In general, don't use `assert` to do argument checking. Use Guava `Preconditions` or something equivalent.
See this page on [assertions] (https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html) for more details

> Do not use assertions for argument checking in public methods.
Argument checking is typically part of the published specifications (or contract) of a method, and these specifications must be obeyed whether assertions are enabled or disabled.

core/src/main/java/io/ddf/content/Schema.java
```diff
@@ -1,655 +1,193 @@
...
-      return null;
-    String[] segments = columnList.split("\\s*,\\s*");
+@NotThreadSafe
+public  class Schema {
 
-    mColumns.clear();
-    for (String segment : segments) {
-      if (Strings.isNullOrEmpty(segment))
-        continue;
+  protected List<String> colNames;
```
sameerp: Why are these `protected`? Do we want subclasses of `Schema`?

sameerp: Why is there are separate list of column names and columns? Why not just one list of columns?

core/src/main/java/io/ddf/content/Schema.java
```diff
@@ -1,655 +1,193 @@
...
-  public Column getColumn(int i) {
-    if (mColumns.isEmpty()) {
-      return null;
-    }
-    if (i < 0 || i >= mColumns.size()) {
-      return null;
-    }
-
-    return mColumns.get(i);
+  public int getColumnIndex(String columnName) throws DDFException {
```
sameerp: This does not throw `DDFException`.

core/src/main/java/io/ddf/content/Schema.java
```diff
@@ -1,655 +1,193 @@
...
   public String toString() {
-    Iterable<String> columnSpecs = Iterables.transform(getColumns(), new Function<Column, String>() {
-      @Override
-      public String apply(Column column) {
-        return column.getName() + " " + column.getType();
+    StringBuilder sb = new StringBuilder();
+    for(Column column : columns){
+      try {
+        sb.append(column.getName() + ":" + column.getType().getSimpleName() + "|");
+      }catch(Exception ex){
```
sameerp: What exception do you expect here?

core/src/main/java/io/ddf/content/Schema.java
```diff
@@ -1,655 +1,193 @@
...
-          temp = new HashMap<String, java.lang.Double>();
-          while (valuesIterator.hasNext()) {
-            String columnValue = valuesIterator.next();
-            temp.put(columnValue, Double.parseDouble(i + ""));
-            i += 1;
-          }
-          dummyCoding.getMapping().put(currentColumnIndex, temp);
-          count += temp.size() - 1;
+      @Override
+      protected Class inferType(String typeName) throws SchemaException {
```
sameerp: Are these the only types we can support at the DDF level? Spark RDDs can contain other arbitrary types. This appears like a departure from our existing implementation.

core/src/main/java/io/ddf/content/Schema.java
```diff
@@ -1,655 +1,193 @@
...
-    this.mTableName = mTableName;
+  public int getNumColumn() {
+    return columns.size();
   }
 
   public List<Column> getColumns() {
-    List<Column> columns = new ArrayList<Column>();
-    for (Column column : mColumns) {
-      columns.add(column.clone());
-    }
```
Huandao0812: getColumns should clone the list of columns and return the cloned list, avoiding user to modify the schema by modifying the columns

PR #281, opened 2016-03-10
remove levelCounts in Factor

Huandao0812:


hai-adatao:
@phvu @nhanitvn @lebinh 
Please help review

Huandao0812:
not done yet

hai-adatao:
This is a big change, please help build a cluster to test when this is done
@lebinh @thxph

hai-adatao:
@nhanitvn @lebinh please help review

core/src/main/java/io/ddf/Factor.java
```diff
@@ -171,38 +206,22 @@ public void setLevels(List<String> levels) throws DDFException {
    *                  meaning
    * @throws DDFException
    */
-  public void setLevels(List<String> levels, List<Integer> codes, boolean isOrdered) throws DDFException {
+  public void setLevels(List<Object> levels, List<Integer> codes, boolean isOrdered) throws DDFException {
```
nhanitvn: Should we just take the list of levels only?

core/src/main/java/io/ddf/content/Schema.java
```diff
@@ -597,7 +527,6 @@ public static ColumnClass get(ColumnType type) {
 
   public static class DummyCoding implements Serializable {
```
nhanitvn: Not the topic of this PR, but we should move DummyCoding out of Schema class into a Pipeline Stage.

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
```diff
@@ -89,18 +82,29 @@ public synchronized void computeFactorLevelsAndLevelCounts() throws DDFException
...
+        listLevelCountsWithName.put(colName, listLevelCounts.get(columnIndex));
       }
+      return listLevelCountsWithName;
+    } else {
+      return new HashMap<String, Map<String, Integer>>();
     }
   }
+  @Override public List<Object> computeFactorLevels(String columnName) throws DDFException {
+    DataFrame df = (DataFrame) this.getDDF().getRepresentationHandler().get(DataFrame.class);
+    DataFrame distinctDF = df.select(columnName).distinct();
```
nhanitvn: Nice, we should use Spark DataFrame as much as possible.

spark/src/main/scala/io/ddf/spark/analytics/FactorIndexer.scala
```diff
@@ -0,0 +1,228 @@
...
+            }
+            arrBuffer += Row.fromSeq(newRow)
+          }
+          arrBuffer.toIterator
+        }
+      }
+    }
+  }
+
+  def inversedTransform(ddf: DDF): DDF = {
```
nhanitvn: Is this for unfactoring?

spark/src/test/java/io/ddf/spark/etl/TransformationHandlerTest.java
```diff
@@ -132,7 +132,7 @@ public void testReservedFactor() throws DDFException {
     ddf.setAsFactor("month");
```
nhanitvn: No need to do in this PR, but we should have a way for users to provide the list of levels in a certain order OR a map from level to index. One example is in binary classification with two labels Yes an No, Yes may be map to 1 as the positive class and vice versa depends on users' need.

nhanitvn: Ah, users can do like this newddf.getSchemaHandler().setAsFactor(column).setLevels(levels);

nhanitvn:
lgtm

core/src/main/java/io/ddf/content/SchemaHandler.java
```diff
@@ -97,68 +101,39 @@ public Schema generateSchema() throws DDFException {
   public synchronized void setFactorLevels(String columnName, Factor<?> factor) throws DDFException {
```
nhanitvn: The name and the param factor of type Factor<?> does not fit together. Should be setFactorLevels(List<Object> levels) ?

core/src/main/java/io/ddf/Factor.java
nhanitvn: OK we have the above setLevels which accept levels only, :)

core/src/main/java/io/ddf/Factor.java
```diff
@@ -99,14 +113,15 @@ public Factor(String name, T[] data) throws DDFException {
...
     super(name, data, engineName);
+    this.mType= type;
   }
 
 
   private Map<String, Integer> mLevelMap;
-  private List<String> mLevels;
-  private Map<String, Integer> mLevelCounts;
+
+  private List<Object> mLevels;
```
nhanitvn: if the mLevelMap is instantiated as a LinkedHashMap which reserves the order of inserted key and thus we can get the list of levels from it with in the original order, why do we need this mLevels list?

core/src/main/java/io/ddf/Factor.java
```diff
@@ -171,38 +206,22 @@ public void setLevels(List<String> levels) throws DDFException {
...
-
-    if (mLevelMap == null) mLevelMap = this.instantiateSynchronizedLevelMap();
-
-    if (codes == null) {
-      // Auto-create a 1-based level-code map
-      codes = new ArrayList<Integer>();
-      for (int i = 1; i <= levels.size(); i++) {
-        codes.add(i);
+    if(codes != null) {
+      if (mLevelMap == null) mLevelMap = this.instantiateSynchronizedLevelMap();
```
lebinh: Even with a synchronized map, this Factor class won't be thread safe. It's mutable, keep it's state in multiple fields and there is no synchronization on all the methods that mutate its state.

core/src/main/java/io/ddf/Factor.java
```diff
@@ -220,4 +239,83 @@ public boolean isOrdered() {
   public void setOrdered(boolean isOrdered) {
     this.mIsOrdered = isOrdered;
   }
+
+  public static class FactorBuilder {
```
lebinh: Why is this builder class is not the only way to initiate the Factor class? It is also not including some fields like `isOrdered`. Looks like we should remove the other public constructors of Factor and make this the only way to create a new Factor? Also we can make Factor an immutable class by moving `computeLevelMap` and `setLevels` from Factor to this builder. That will make Factor truly thread safe.

core/src/main/java/io/ddf/Factor.java
```diff
@@ -99,14 +113,15 @@ public Factor(String name, T[] data) throws DDFException {
...
    * @throws DDFException
    */
-  public Factor(String name, T[] data, String engineName) throws DDFException {
+  public Factor(String name, T[] data, String engineName, Schema.ColumnType type) throws DDFException {
     super(name, data, engineName);
+    this.mType= type;
   }
 
 
   private Map<String, Integer> mLevelMap;
```
lebinh: Why `mLevels` is a `List<Object>` but `mLevelMap` is a map with `String` as key?

core/src/main/java/io/ddf/content/SchemaHandler.java
lebinh: Agree, if this only need a levels list then let's not use a Factor as a param

lebinh:
@Huandao0812 can we have some high level overview of the change here? It's hard navigating through the changes without that.

ddf-test/src/main/scala/io/ddf/test/it/SchemaSpec.scala
```diff
@@ -91,65 +92,28 @@ trait SchemaSpec extends BaseSpec with Matchers {
...
       }
-      schemaHandler.computeFactorLevelsAndLevelCounts()
+      val factorMap = schemaHandler.computeLevelCounts(columnNames)
 
       val cols = Array(0, 8, 16, 17, 24, 25).map {
         idx => schemaHandler.getColumn(schemaHandler.getColumnName(idx))
       }
-      assert(cols(0).getOptionalFactor.getLevels.contains("2008"))
-      assert(cols(0).getOptionalFactor.getLevels.contains("2010"))
-      assert(cols(0).getOptionalFactor.getLevelCounts.get("2008") === 28.0)
```
hai-adatao: Some assertions are not migrated (from cols0 - cols2)

ddf-test/src/main/scala/io/ddf/test/it/SchemaSpec.scala
```diff
@@ -91,65 +92,28 @@ trait SchemaSpec extends BaseSpec with Matchers {
...
-      assert(cols(3).getOptionalFactor.getLevelCounts.get("MCO") === 3.0)
-      assert(cols(3).getOptionalFactor.getLevelCounts.get("TPA") === 3.0)
-      assert(cols(3).getOptionalFactor.getLevelCounts.get("JAX") === 1.0)
-      assert(cols(3).getOptionalFactor.getLevelCounts.get("LAS") === 3.0)
-      assert(cols(3).getOptionalFactor.getLevelCounts.get("BWI") === 10.0)
-
-      assert(cols(5).getOptionalFactor.getLevelCounts.get("0") === 9.0)
-      assert(cols(4).getOptionalFactor.getLevelCounts.get("3") === 1.0)
-
-      val ddf2 = manager.sql2ddf("select * from airlineWithNA", engineName)
```
hai-adatao: Tests with NA is removed

spark/src/test/scala/io/ddf/spark/content/FactorSuite.scala
```diff
@@ -138,17 +80,18 @@ class FactorSuite extends ATestSuite {
     val schemaHandler = ddf.getSchemaHandler
     val factorColumns = Array("vs", "am", "gear", "carb")
     factorColumns.foreach {
-      column => schemaHandler.setAsFactor(column)
+      column =>
+        val factor = schemaHandler.setAsFactor(column)
+//        factor.setLevels(schemaHandler.computeFactorLevels(column))
```
hai-adatao: If not used, remove instead of comment out

spark/src/test/scala/io/ddf/spark/content/FactorSuite.scala
```diff
@@ -138,17 +80,18 @@ class FactorSuite extends ATestSuite {
...
     }
-    schemaHandler.computeFactorLevelsAndLevelCounts()
+//    schemaHandler.computeLevelCounts(factorColumns)
```
hai-adatao: here too

ddf-test/src/main/scala/io/ddf/test/it/SchemaSpec.scala
Huandao0812: duplicated test

core/src/main/java/io/ddf/Factor.java
Huandao0812: will fix in another PR

core/src/main/java/io/ddf/Factor.java
Huandao0812: another PR

core/src/main/java/io/ddf/Factor.java
Huandao0812: another PR

spark/src/main/scala/io/ddf/spark/analytics/FactorIndexer.scala
Huandao0812: no, it's to inversedTransform the factor column, different from unfactor

hai-adatao:
I'm gonna merge this, @Huandao0812 fix help create tickets for later fix / improvement if necessary

PR #286, opened 2016-03-22
Provide more appropriate error messages for groupby and aggregate APIs

nhanitvn:


spark/src/main/scala/io/ddf/spark/analytics/AggregationHandler.scala
```diff
@@ -0,0 +1,40 @@
...
+ * Created by nhanitvn on 3/21/16.
+ */
+class AggregationHandler(theDDF: DDF) extends CoreAggregationHandler(theDDF) {
+  override def groupBy(groupedColumns: JList[String], aggregateFunctions: JList[String]): DDF = {
+    // Catch SparkSQL/JSQLParser error message, parse and produce appropriate error message
+    try {
+      super.groupBy(groupedColumns, aggregateFunctions)
+    } catch {
+      case ddfException: DDFException => throw new DDFException(SparkUtils.sqlErrorToDDFError(ddfException.getMessage, buildGroupBySQL(aggregateFunctions), aggregateFunctions.asScala.toList:::groupedColumns.asScala.toList))
+      case throwable: Throwable => throw throwable
```
lebinh: This case clause is unnecessary. You don't have to catch if you just throw it again as that's the default behaviours.

spark/src/main/scala/io/ddf/spark/analytics/AggregationHandler.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  override def aggregate (fields: JList[AggregateTypes.AggregateField]): AggregateTypes.AggregationResult = {
+    // Catch SparkSQL/JSQLParser error message, parse and produce appropriate error message
+    try {
+      super.aggregate(fields)
+    } catch {
+      case ddfException: DDFException =>
+        val tableName: String = this.getDDF.getTableName
+        throw new DDFException(SparkUtils.sqlErrorToDDFError(ddfException.getMessage, AggregateField.toSql(fields, tableName), fields.asScala.toList.map(_.toString)))
+      case throwable: Throwable => throw throwable
```
lebinh: Same as above for catch and throw

core/src/main/java/io/ddf/analytics/AggregationHandler.java
```diff
@@ -122,24 +122,31 @@ public DDF groupBy(List<String> groupedColumns) {
     return this.getDDF();
   }
 
+  protected String buildGroupBySQL(List<String> aggregateFunctions) {
+    String groupedColSql = Joiner.on(",").join(mGroupedColumns);
+
+    String selectFuncSql = convertAggregateFunctionsToSql(aggregateFunctions.get(0));
+    for (int i = 1; i < aggregateFunctions.size(); i++) {
+      selectFuncSql += "," + convertAggregateFunctionsToSql(aggregateFunctions.get(i));
```
lebinh: Why not using Joiner as above as it is way clearer? With map then join?

nhanitvn: Good point. This is the legacy code. I just moved it out of the groupby() function for usage in the other places. Let me take a look based on your comment, :)

spark/src/main/scala/io/ddf/spark/util/SparkUtils.scala
```diff
@@ -286,4 +286,70 @@ object SparkUtils {
...
+    val InvalidCharacter2 = ".*character '(.)' not supported here;.*".r
+
+    // non-existent column SparkSQL
+    // "cannot resolve 'substring' given input columns ; line 1 pos 7"
+    val NonexistentColumns = ".*cannot resolve '(.+)' given input columns.*".r
+
+    def handleInvalidCharacter(sqlError: String, char: String): String = {
+      var errorMessage = s"Expressions or columns containing invalid character $char: "
+      val exprs = expressions.filter(exp => exp.contains(char))
+      if (exprs.nonEmpty) {
```
lebinh: This is redundant as `mkString` on empty list will just return empty string and thus the `errorMessage` will be the same. We can save 3 line here :smile:

spark/src/main/scala/io/ddf/spark/util/SparkUtils.scala
```diff
@@ -286,4 +286,70 @@ object SparkUtils {
...
+
+    def handleInvalidExpression(sqlError: String, pos: Int): String = {
+      // Find the expression right before the position
+      val exprPos = expressions.map(exp => sql.substring(0, pos).indexOf(exp)).zipWithIndex.maxBy(_._1)
+      s"Column or Expression with invalid syntax: '${expressions(exprPos._2)}'"
+    }
+
+    def handleGeneralSQLError(sqlError: String): String = {
+      // Remove Exception info
+      val sqlErrorRemovedException = sqlError.substring(sqlError.indexOf("Exception:")+11, sqlError.length)
```
lebinh: Magic number!

spark/src/main/scala/io/ddf/spark/util/SparkUtils.scala
nhanitvn: I removed a logic and forget to update the code, ha ha.

core/src/main/java/io/ddf/analytics/AggregationHandler.java
nhanitvn: Oh, this is Java code. Do we completely switch to Java 8 now?

lebinh: Not completely, for now we can use Guava's List.transform (i.e. Scala map), http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/Lists.html#transform(java.util.List,%20com.google.common.base.Function)

lebinh: It will look uglier than Scala or Java8 though :(

core/src/main/java/io/ddf/analytics/AggregationHandler.java
```diff
@@ -122,31 +103,37 @@ public DDF groupBy(List<String> groupedColumns) {
...
 
+  protected String buildGroupBySQL(List<String> aggregateFunctions) throws DDFException {
+    String groupedColSql = Joiner.on(",").join(mGroupedColumns);
+
+    List<String> aggregationFuncSql = new ArrayList<String>();
+    for (int i = 0; i < aggregateFunctions.size(); i++) {
+      aggregationFuncSql.add(convertAggregateFunctionsToSql(aggregateFunctions.get(i)));
+    }
+
+    String selectFuncSql = Joiner.on(",").join(aggregateFunctions);
```
nhanitvn: @lebinh That Google Function does not worth to use here in Java code. I refactor a bit for easier understanding. Comment? :)

lebinh: That's ok but please use for each loop instead of for by index

spark/src/test/scala/io/ddf/spark/analytics/AggregationHandlerSuite.scala
```diff
@@ -0,0 +1,92 @@
...
+    val ddf = manager.sql2ddf("select * from airline", "SparkSQL").asInstanceOf[SparkDDF]
+
+    val thrown1 = intercept[DDFException]{ddf.groupBy(List("Year").asJava, List("newcol@=avg(arrdelay)").asJava)}
+    assert(thrown1.getMessage === "Expressions or columns containing invalid character @: newcol@=avg(arrdelay)")
+  }
+
+  test("Proper error message for new columns that exist") {
+    val ddf = manager.sql2ddf("select * from airline", "SparkSQL").asInstanceOf[SparkDDF]
+
+    val thrown1 = intercept[DDFException]{ddf.groupBy(List("Year").asJava, List("Year=avg(arrdelay)").asJava)}
```
khangich: @nhanitvn @binhle the ddf.groupby return a new ddf. And in this new ddf it's entirely ok to have column named Year, arrdelay. 

Because it's new ddf. Why do we have to check if column is existed ?

lebinh: @khangich the check here is correct as the new DDFs have `Year` for both 2 column name, but your case is a bug, fixing

PR #287, opened 2016-03-23
Fix group by checking for duplicated column name with original DDF

lebinh:
This fix https://adatao.atlassian.net/browse/PE-1716

Checking for duplicated column names on aggregation functions and group by columns instead of columns of the source DDF.

Reviewer: @nhanitvn

### DIFF
--- a/core/src/main/java/io/ddf/analytics/AggregationHandler.java
+++ b/core/src/main/java/io/ddf/analytics/AggregationHandler.java
@@ -13,7 +13,9 @@ import io.ddf.types.AggregateTypes.AggregationResult;
 import io.ddf.util.Utils;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 /**
  *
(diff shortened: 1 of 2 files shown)


lebinh:
@nhanitvn I self merge this as it is kind of urgent but please review when you wake up

core/src/main/java/io/ddf/analytics/AggregationHandler.java
```diff
@@ -146,21 +158,46 @@ public DDF agg(List<String> aggregateFunctions) throws DDFException {
     }
   }
 
-  private String convertAggregateFunctionsToSql(String sql) throws DDFException {
+  private static class AggregatedColumnExpression {
```
nhanitvn: Nice refactoring, :)

core/src/main/java/io/ddf/analytics/AggregationHandler.java
```diff
@@ -146,21 +158,46 @@ public DDF agg(List<String> aggregateFunctions) throws DDFException {
...
+    }
+
+    // TODO: this function need test
+    public static AggregatedColumnExpression fromString(String exp) throws DDFException {
+      if (Strings.isNullOrEmpty(exp)) {
+        throw new DDFException("Aggregation function cannot be null or empty");
+      }
 
-    if (Strings.isNullOrEmpty(sql)) return null;
+      // XXX @nhanitvn what is this regexp doing?
```
nhanitvn: A legacy code. I am working on expression processing for transform API too. Let's me handle this in that PR too.

PR #290, opened 2016-03-29
Add applySchema to DDF

zkidkid:
### Description
Add new feature "ApplySchema", this function will allow to change column types from string to other for data-ingestion.
Reviewers: @sameerp @mbbui @PangZhi @nhanitvn @Huandao0812 
P/S: Please help to review it carefully, because I'm not familar with Spark so maybe my approach is not efficient. 

### How to test
UnitTest
Cmd: mvn surefire:test -Dtest=SchemaHandlerTest
or run SchemaHandlerTest directly from IDEA.

### PR Progress
- [ ] Code review is done.


spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
```diff
@@ -106,5 +127,260 @@ public static Schema getSchemaFromDataFrame(DataFrame rdd) {
...
+    }
+    return listConvertColumn;
+  }
+
+  /**
+   * ApplySchemaStatistic will hold: + total line processed + total line success + column failed parsed info (total
+   * failed on this column & example )
+   */
+  @NotThreadSafe
+  public static class ApplySchemaStatistic implements Serializable {
```
Huandao0812: why do we need this?

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
```diff
@@ -106,5 +127,260 @@ public static Schema getSchemaFromDataFrame(DataFrame rdd) {
...
+   * new DDF and also statistic when its applied new schema
+   *
+   * @param schema will use to apply for current DDF
+   * @return Tuple2<SparkDDF,ApplySchemaStatistic> +SparkDDF: new sparkddf which applied new schema already
+   * +ApplySchemaStatistic:  applied schema statistic @see ApplySchemaStatistic
+   */
+  public Tuple2<SparkDDF, ApplySchemaStatistic> applySchema(Schema schema) throws DDFException {
+    long startApplySchema = System.currentTimeMillis();
+    // Keep info to Convert column at <pos> to <ColumnType>
+    List<Tuple3<Integer, String, Schema.ColumnType>> listConvertColumns = getListConvertColumn(this.getSchema(), schema);
```
Huandao0812: bad variable name

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
```diff
@@ -106,5 +127,260 @@ public static Schema getSchemaFromDataFrame(DataFrame rdd) {
...
+
+    try {
+      SparkDDFManager manager = (SparkDDFManager) this.getManager();
+//      SparkContext sparkContext = manager.getSparkContext();
+      Accumulator<ApplySchemaStatistic> statisticAccumulator = manager.getSparkContext().accumulator(new ApplySchemaStatistic(), ApplySchemaStatisticParam.getInstance());
+
+      JavaRDD<Row> rdd = ((SparkDDF) this.getDDF()).getRDD(Row.class).toJavaRDD();
+      JavaRDD<Row> filteredRdd = rdd.filter(row -> convertRowType(row, listConvertColumns, statisticAccumulator));
```
Huandao0812: it's better to use DataFrame api to change schema, using RDD then map over rows is rather tedious and error prone.
See https://spark.apache.org/docs/1.6.0/api/java/org/apache/spark/sql/Column.html#cast(org.apache.spark.sql.types.DataType)

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
zkidkid: could you make your question more clearly @Huandao0812

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
zkidkid: You mean listConvertColumns right ? Could you help to suggest a new one.

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
```diff
@@ -106,5 +127,260 @@ public static Schema getSchemaFromDataFrame(DataFrame rdd) {
...
+
+  /**
+   * Check whether we could convert data at row[index] requiredType @see Schema.ColumnType
+   *
+   * @param row:          Which row to convert
+   * @param index:        index of column in row to convert
+   * @param requiredType: type we need to convert to
+   * @return whether we could convert row[index] to requiredType
+   */
+  private boolean isConvertible(Row row, Integer index, Schema.ColumnType requiredType) {
```
Huandao0812: Spark DataFrame and DDF already have ColumnType, checking isConvertible at row level will be very slow.

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
```diff
@@ -106,5 +127,260 @@ public static Schema getSchemaFromDataFrame(DataFrame rdd) {
...
+
+    try {
+      if(row.get(index) == null) return false;
```
Huandao0812: also use row.isNullAt(index) to check for null

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
```diff
@@ -106,5 +127,260 @@ public static Schema getSchemaFromDataFrame(DataFrame rdd) {
...
+      String columnData = row.get(index).toString().trim();
```
Huandao0812: SparkSQL Row has api to get row's primitive values, e.g: row.getInt(index), row.getLong(index), etc...
see https://spark.apache.org/docs/1.4.0/api/java/org/apache/spark/sql/Row.html

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
```diff
@@ -106,5 +127,260 @@ public static Schema getSchemaFromDataFrame(DataFrame rdd) {
...
+   * Check whether we could convert data at row[index] requiredType @see Schema.ColumnType
+   *
+   * @param row:          Which row to convert
+   * @param index:        index of column in row to convert
+   * @param requiredType: type we need to convert to
+   * @return whether we could convert row[index] to requiredType
+   */
+  private boolean isConvertible(Row row, Integer index, Schema.ColumnType requiredType) {
+
+    try {
```
Huandao0812: should avoid try catch at this tight loop, as it's very expensive

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
Huandao0812: also there's no reason null is not convertible, for example if the value is null in original type, it should be simply null in the new type

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
```diff
@@ -106,5 +127,260 @@ public static Schema getSchemaFromDataFrame(DataFrame rdd) {
...
+    }
+    return false;
+  }
+
+  /**
+   * get list column need to be convert
+   *
+   * @return List<Tuple3<Integer,String,ColumnType>> = ColumnIndex,ColumnName,ColumnType
+   */
+  protected List<Tuple3<Integer, String, Schema.ColumnType>> getListConvertColumn(Schema fromSchema, Schema toSchema) throws DDFException {
```
Huandao0812: if you use Spark DataFrame api to cast, then don't need the index of column anymore, also Tuple2<String, ColumnType> should simply be Column

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
zkidkid: I think it will do type cast (dynamic cast) so it will throw ClassCastException instead.

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
zkidkid: Hi @Huandao0812 ,
I don't know how to do in DataFrame, let I check. 
Currently my approach is using Rdd<Row> instead.
Do you know whether DataFrame could give out the statistic ?

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
zkidkid: it should be an option from user. will update later.

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
Huandao0812: it will not throw ClassCastException if you use the corresponding api for the columnType,
DDF has the column type information, so for index = 1, you should know what is the correct api to call.

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
zkidkid: do you mean this code @Huandao0812 :` df.select(df("colA").cast("int"))`
So how do we get statistic for sample data & how many row failed on this column ?
btw, what is the reason to make this code slow? I think it's also O(1) ?

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
zkidkid: Column type is String @Huandao0812 .

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
Huandao0812: I mean why do we need the statistics?

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
Huandao0812: ah, sorry, missed the use case, it to cast from String to respective type

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
zkidkid: try here is to know whether casting to expected type is possible. what is your solution to get over this ?

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
Huandao0812: using DataFrame api to cast could by simply like this 
```
val dataFrame = ddf.getRepresentationHandler.get(classOf[DataFrame]).asInstanceOf[DataFrame]
    val newDF = dataFrame.select(dataFrame.col("vs"), dataFrame.col("gear"), dataFrame.col("mpg").cast("double"), dataFrame.col("am").cast("string"))
```

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
zkidkid: kk I also don't know :v 
You could check it here: https://docs.google.com/document/d/1dSw-4zPXq5HQ99-VO1Pt0fyb58slRus8wZnlzp90ftI/edit#heading=h.fm8rwtgx02uq

zkidkid: if you lazy to read, here is summary for statistic:
1/ how may line succeed/line processed 
2/ for every column, how many time it failed & give some example.
 ```
   "dataContainerID": "ddf_uuid"
    "errors": {
        "column_a": {
            "dropped": {
                "count": 180,
                "examples": ["123def", "456abc"]
            },
```
If you still feel confused, let read the doc or call TuanAnh.

File: spark/src/main/java/io/ddf/spark/content/SchemaHandler.java (hunk 3 above)
Comment (zkidkid) [inline spark/src/main/java/io/ddf/spark/content/SchemaHandler.java]: could you read my function description:
```
+  /**
 +   * applySchema(Schema newSchema) will do apply change column type from current to new data type. This function return
 +   * new DDF and also statistic when its applied new schema
 +   *
 +   * @param schema will use to apply for current DDF
 +   * @return Tuple2<SparkDDF,ApplySchemaStatistic> +SparkDDF: new sparkddf which applied new schema already
 +   * +ApplySchemaStatistic:  applied schema statistic @see ApplySchemaStatistic
 +   */
 +  public Tuple2<SparkDDF, ApplySchemaStatistic> applySchema(Schema schema) throws DDFException {
```

File: spark/src/main/java/io/ddf/spark/content/SchemaHandler.java (hunk 7 above)
Comment (Huandao0812) [inline spark/src/main/java/io/ddf/spark/content/SchemaHandler.java]: see http://stackoverflow.com/questions/1102891/how-to-check-if-a-string-is-numeric-in-java 
another common approach is using regex to check

Comment (zkidkid) [inline spark/src/main/java/io/ddf/spark/content/SchemaHandler.java]: One of problem with regex is it's not well test and doesn't detect overflow.
I would agree that use exception to handle this is not a good solution, but in this situation I will use it first.
In the future, I think we should approach using Spark Parser instead.

File: spark/src/main/java/io/ddf/spark/content/SchemaHandler.java [hunk 9]
```diff
@@ -106,5 +127,260 @@ public static Schema getSchemaFromDataFrame(DataFrame rdd) {
...
+  }
+
+  /**
+   * Convert Row Data To Given Type If failed, store data to statistic
+   *
+   * @param row                  which hold multiple columns to convert
+   * @param listConvertColumns   given columnIndex & columnName &  type to convert to
+   * @param statisticAccumulator statistic
+   */
+  private Boolean convertRowType(Row row, List<Tuple3<Integer, String, Schema.ColumnType>> listConvertColumns, Accumulator<ApplySchemaStatistic> statisticAccumulator) {
```
Huandao0812: this function does not do any conversion yet, it's just checking whether you can convert a row or not.

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
```diff
@@ -106,5 +127,260 @@ public static Schema getSchemaFromDataFrame(DataFrame rdd) {
...
+    boolean result = true;
+    statisticAccumulator.localValue().increaseLineProcessed();
+    for (Tuple3<Integer, String, Schema.ColumnType> tuple3 : listConvertColumns) {
+      if (!isConvertible(row, tuple3._1(), tuple3._3())) {
+        statisticAccumulator.localValue().addColumnFailed(tuple3._2(), row.getString(tuple3._1()));
+        result = false;
```
Huandao0812: why would a row fail to be converted if just one of the value fail?

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
Huandao0812: why do we use rdd.filter here? if you want to convert elements of rdd, must use rdd.map instead

spark/src/test/java/io/ddf/spark/content/SchemaHandlerTest.java
```diff
@@ -0,0 +1,69 @@
...
+    columns.add(new Schema.Column("type_short", Schema.ColumnType.SMALLINT));
+    columns.add(new Schema.Column("type_long", Schema.ColumnType.BIGINT));
+    columns.add(new Schema.Column("type_double", Schema.ColumnType.DOUBLE));
+    columns.add(new Schema.Column("type_decimal", Schema.ColumnType.DECIMAL));
+    columns.add(new Schema.Column("type_date", Schema.ColumnType.DATE));
+    columns.add(new Schema.Column("type_timestamp", Schema.ColumnType.TIMESTAMP));
+
+    Schema schema = new Schema(columns);
+    Tuple2<SparkDDF, SchemaHandler.ApplySchemaStatistic> applySchema = schemaHandler.applySchema(schema);
+    SparkDDF sparkDDF = applySchema._1();
```
Huandao0812: should check schema of the new DDF

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
Huandao0812: it's perfectly fine to have null values in DDF/DataFrame, if a value in a row failed to convert to new type, just put the corresponding value in the new row to null.

spark/src/main/java/io/ddf/spark/content/SchemaHandler.java
PangZhi: @Huandao0812 @zkidkid I think the pain point here is that we not only want to do the type conversion, but also wants to get statistics for the conversion. For example, how many rows fails. Thus we need things like counter. But i'm also concerned about the performance, although both may be O(1), but still performance can be different. @zkidkid Do you think we should do a small benchmark using spark to check the performance of scanning row by row and dataframe's cast?

PR #300, opened 2016-04-06
Separate New column names and transform expressions in TransformSQL

phvu:
- Added transformUDFWithNames to replace transformUDF. Separating names and transformation expressions
- Deprecated transformUDF (in `IHandleTransformations`)

Reviewers: @Huandao0812 @nhanitvn @hai-adatao 

No

### How to test
UT

### PR Progress
- [x] Code review is done.

### DIFF
--- a/core/src/main/java/io/ddf/etl/TransformationHandler.java
+++ b/core/src/main/java/io/ddf/etl/TransformationHandler.java
@@ -1,7 +1,7 @@
 package io.ddf.etl;
 
 
-import com.google.common.base.Joiner;
+
 import com.google.common.collect.Lists;
 import io.ddf.DDF;
 import io.ddf.analytics.Summary;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -222,14 +218,131 @@ public static String RToSqlUdf(List<String> RExps, List<String> selectedColumns,
...
       selectStr += (exp + ",");
     }
 
     return selectStr.substring(0, selectStr.length() - 1);
   }
 
+  /**
+   * Create new columns or overwrite existing ones
+   *
+   * @param newColumnNames       Array of new column names.
```
nhanitvn: Should add details on the case of anonymous expressions

phvu: done

Huandao0812: can we limit the number of column and expression to be one at a time, I don't see many use case where DS need to transform multiple columns at a time.

phvu: Maybe no in interactive scenarios, but once they write it up in a script, they may want to pack them all into a single function invoke.
Is there any technical disadvantage with this approach though? We can add wrapper which takes single expression, similar to `transformUDF`.

PR #310, opened 2016-04-15
Make transformUDFWithNames & castType & removeColumns return new DDF by default

Huandao0812:
... with option to modify the current DDF inplace

Reviewers: @phvu @nhanitvn @zkidkid @PangZhi 

Change in transformUDFWithNames & castType behaviour

### PR Progress
- [ ] Code review is done.

### DIFF
--- a/core/src/main/java/io/ddf/content/ViewHandler.java
+++ b/core/src/main/java/io/ddf/content/ViewHandler.java
@@ -107,6 +107,17 @@ public class ViewHandler extends ADDFFunctionalGroupHandler implements IHandleVi
   }
 
   @Override
+  public DDF removeColumn(String columnName, Boolean inPlace) throws DDFException {
+    DDF newDDF = this.removeColumn(columnName);
+    if(inPlace) {
+      this.getDDF().getMutabilityHandler().updateInplace(newDDF);
+      return this.getDDF();
+    } else {
+      return newDDF;
+    }
+  }
+
+  @Override
   public DDF removeColumns(String... columnNames) throws DDFException {
     if (columnNames == null || columnNames.length == 0) throw new DDFException("columnNames must be specified");
 
@@ -126,19 +137,29 @@ public class ViewHandler extends ADDFFunctionalGroupHandler implements IHandleVi
         throw new DDFException(String.format("Column %s does not exists", columnName));
       }
     }
-    List<String> columns = this.getDDF().getColumnNames();
 
-    for (String columnName : currentColumnNames) {
-      for (Iterator<String> it = columns.iterator();it.hasNext();) {
+    for (String columnName : columnNames) {
+      for (Iterator<String> it = currentColumnNames.iterator();it.hasNext();) {
         if (it.next().equals(columnName)) {
           it.remove();
         }
       }
     }
 
-    DDF newddf = this.project(columns);
+    DDF newddf = this.project(currentColumnNames);
     newddf.getMetaDataHandler().copyFactor(this.getDDF());
     return newddf;
+  }
+
+  @Override
+  public DDF removeColumns(List<String> columnNames, Boolean inPlace) throws DDFException {
+    DDF newDDF = this.removeColumns(columnNames);
+    if(inPlace) {
+      this.getDDF().getMutabilityHandler().updateInplace(newDDF);
+      return this.getDDF();
+    } else {
+      return newDDF;
+    }
   }
 
   // ///// Execute SQL command on the DDF ///////
--- a/core/src/main/java/io/ddf/content/ViewHandler.java
+++ b/core/src/main/java/io/ddf/content/ViewHandler.java
@@ -128,7 +128,7 @@ public class ViewHandler extends ADDFFunctionalGroupHandler implements IHandleVi
     }
     List<String> columns = this.getDDF().getColumnNames();
 
-    for (String columnName : columnNames) {
+    for (String columnName : currentColumnNames) {
       for (Iterator<String> it = columns.iterator();it.hasNext();) {
         if (it.next().equals(columnName)) {
           it.remove();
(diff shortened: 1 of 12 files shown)


hai-adatao:
Is this in 1.4.14? This looks like a big breaking change

core/src/main/java/io/ddf/etl/IHandleTransformations.java
```diff
@@ -44,12 +44,29 @@ DDF transformPython(String[] transformFunctions, String[] functionNames,
    *                             with newColumnNames.
    * @param selectedColumns list of column names to be included in the result DDF.
    *                        If null or empty, all existing columns will be included.
-   * @return current DDF if it is mutable, or a new DDF otherwise.
+   * @return new DDF
    * @throws DDFException
    */
   DDF transformUDFWithNames(String[] newColumnNames, String[] transformExpressions,
```
PangZhi: One question is: why we didn't do case-insensitive check here? @Huandao0812 @hai-adatao This function only did case sensitive check here.

PangZhi:
@hai-adatao @Huandao0812 i think we need to discuss about the case-insensitive behavior.

core/src/main/java/io/ddf/content/ViewHandler.java
```diff
@@ -128,7 +128,7 @@ public DDF removeColumns(List<String> columnNames) throws DDFException {
     }
     List<String> columns = this.getDDF().getColumnNames();
```
hai-adatao: This is not different from `List<String> currentColumnNames = this.getDDF().getColumnNames();`, why declare again?

core/src/main/java/io/ddf/content/ViewHandler.java
```diff
@@ -128,7 +128,7 @@ public DDF removeColumns(List<String> columnNames) throws DDFException {
...
 
-    for (String columnName : columnNames) {
+    for (String columnName : currentColumnNames) {
```
hai-adatao: The old implementation is correct

core/src/main/java/io/ddf/content/ViewHandler.java
```diff
@@ -137,13 +137,8 @@ public DDF removeColumns(List<String> columnNames) throws DDFException {
...
 
     DDF newddf = this.project(columns);
-    if(this.getDDF().isMutable()) {
-      this.getDDF().updateInplace(newddf);
-      return this.getDDF();
-    } else {
-      newddf.getMetaDataHandler().copyFactor(this.getDDF());
-      return newddf;
-    }
+    newddf.getMetaDataHandler().copyFactor(this.getDDF());
```
hai-adatao: Need to note this in the description, current behavior of remove_columns in PyClient is in place, this will break current behavior

core/src/main/java/io/ddf/content/ViewHandler.java
phvu: `ViewHandler` was changed by @Huandao0812, I didn't change anything there:
https://github.com/ddf-project/DDF/commits/feature/immutable/core/src/main/java/io/ddf/content/ViewHandler.java

I noticed something crazy with removeColumns() on this branch too. Will take a look later, unless @Huandao0812 wanna look into this first.

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -361,12 +343,18 @@ public synchronized DDF transformUDFWithNames(String[] newColumnNames, String[]
 
     DDF newddf = this.getManager()
         .sql2ddf(sqlCmd, new SQLDataSourceDescriptor(sqlCmd, null, null, null, this.getDDF().getUUID().toString()));
+    newddf.getMetaDataHandler().copyFactor(this.getDDF());
```
hai-adatao: 😅 move the implementation to `transformUDFWithNames` with `inPlace` flag and in this method call to that with `inPlace = false` is more natural IMO

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -391,4 +379,17 @@ public static String RToSqlUdf(String RExp) {
   @Override public DDF oneHotEncoding(String inputColumn, String outputColumnName) throws DDFException {
     throw new UnsupportedOperationException();
   }
+
+  @Override public DDF castType(String column, String newType) throws DDFException {
+    throw new UnsupportedOperationException();
```
hai-adatao: Really? This is not `castType` with `inPlace = false`?

hai-adatao: Ah... I understand... it's better to make this method abstract rather than method with inPlace flag

core/src/main/java/io/ddf/content/ViewHandler.java
Huandao0812: not sure what do you mean, of course this will break current behaviour, and we need to change the client also

core/src/main/java/io/ddf/content/ViewHandler.java
Huandao0812: the logic for this function is weird, need to review

core/src/main/java/io/ddf/content/ViewHandler.java
hai-adatao: I mean this PR said it just change `castType` and `transform`, not `removeColumns`
This change will make removeColumns return new DDF instead of the same DDF as of now

core/src/main/java/io/ddf/etl/IHandleTransformations.java
hai-adatao: How about now @PangZhi ?

core/src/main/java/io/ddf/content/ViewHandler.java
hai-adatao: Also, https://ci.arimo.com/job/pull-request-ddf-unittest/io.ddf$ddf_spark_2.10/232/testReport/junit/io.ddf.spark.content/ViewHandlerTest/testRemoveColumns/

testRemoveColumns failed

PR #327, opened 2016-05-03
Can get a DDF by sampling a Spark DDF by size

nhanitvn:
Also:
- The resulted DDF has exact 'size' number of rows
- Sample by fraction now return exact number of rows as specified by fraction
- The old sample by fraction is renamed to indicate that the result is approximated.

Reviewers: @Huandao0812 @phvu @hai-adatao 

No breaking changes in API. However, we need to take care about performance.

### How to test

### PR Progress
- [ ] Code review is done. 

Change-Id: I853d48575247d9179936d2a8c67ae79035ccb959


core/src/main/java/io/ddf/content/IHandleViews.java
```diff
@@ -24,12 +24,14 @@
    */
   public List<Object[]> getRandomSample(int numSamples, boolean withReplacement, int seed);
 
-  public DDF getRandomSampleByNum(int numSamples,
+  public DDF getRandomSampleByNum(long numSamples,
```
phvu: Does this break lineage history?

Huandao0812: we don't have lineage for sample operation anyway

spark/src/main/scala/io/ddf/spark/content/ViewHandler.scala
```diff
@@ -51,10 +56,50 @@ class ViewHandler(mDDF: DDF) extends io.ddf.content.ViewHandler(mDDF) with IHand
...
+
+      // Get an RDD of rows, create a new SparkDF and then a DDF
+      val sampledRDD: RDD[Row] = joinedRDD.map(x => x._2._2.get)
+      val df: DataFrame = mDDF.getRepresentationHandler.get(classOf[DataFrame]).asInstanceOf[DataFrame]
+      val sampledDF = sqlContext.createDataFrame(sampledRDD, df.schema)
+
+
+      val manager = this.getManager
+      val schema = SchemaHandler.getSchemaFromDataFrame(sampledDF)
+      schema.setTableName(mDDF.getSchemaHandler.newTableName())
```
Huandao0812: you don't need to do this, also we have function newDDFFromSparkDataFrame in SparkDDFManager to create DDF from DataFrame

spark/src/main/scala/io/ddf/spark/content/ViewHandler.scala
```diff
@@ -51,10 +56,50 @@ class ViewHandler(mDDF: DDF) extends io.ddf.content.ViewHandler(mDDF) with IHand
...
+      val sampleDDF = manager.newDDF(manager, sampledDF, Array
+      (classOf[DataFrame]), manager.getNamespace,
+        null, schema)
+
+      mLog.info(">>>>>>> adding ddf to DDFManager " + sampleDDF.getName)
+      this.getDDF.getSchemaHandler.getColumns.foreach{
```
Huandao0812: we have copyFactor function in MetaDataHandler

spark/src/main/scala/io/ddf/spark/content/ViewHandler.scala
```diff
@@ -51,10 +56,50 @@ class ViewHandler(mDDF: DDF) extends io.ddf.content.ViewHandler(mDDF) with IHand
...
+
+    if (numSamples < 0) {
+      throw new IllegalArgumentException("Number of samples must be larger than or equal to 0")
+    }
+
+    if (!withReplacement){
+      mDDF.getSqlHandler.sql2ddf(s"select * from ${mDDF.getSchema.getTableName} order by rand($seed) limit $numSamples")
+    } else {
+      // First, create a Spark DF that contain random numbers from 0 to num_rows - 1
+      val numRows = mDDF.getNumRows
```
phvu: Another way to do this is
```
rddRow.sample(true, 2.0*numSamples/rddRow.count(), seed).zipWithIndex().filter(_._2 < numSamples).map(_._1)
```
i.e. we use Spark's API to sample twice what we need, then only pick the first `numSamples` rows.
Not sure if it is more efficient though.

core/src/main/java/io/ddf/content/IHandleViews.java
phvu: then maybe let's change the function names? Those names sound a bit too verbose. Probably something like this:

- `getRandomSample` -> `sample(double fraction, boolean replacement, int seed)`
- `getRandomSampleByNum` -> `sample(long size, boolean replacement, int seed)`
- `getRandomSampleApprox` -> `sampleApprox(double fraction, boolean replacement, int seed)`

nhanitvn: good point. How do you think? @Huandao0812

spark/src/main/scala/io/ddf/spark/content/ViewHandler.scala
```diff
@@ -51,10 +56,50 @@ class ViewHandler(mDDF: DDF) extends io.ddf.content.ViewHandler(mDDF) with IHand
...
+
+
+      val manager = this.getManager
+      val schema = SchemaHandler.getSchemaFromDataFrame(sampledDF)
+      schema.setTableName(mDDF.getSchemaHandler.newTableName())
+      val sampleDDF = manager.newDDF(manager, sampledDF, Array
+      (classOf[DataFrame]), manager.getNamespace,
+        null, schema)
+
+      mLog.info(">>>>>>> adding ddf to DDFManager " + sampleDDF.getName)
```
Huandao0812: this is not necessary, also sampleDDF  does not have name here, only uuid

spark/src/main/scala/io/ddf/spark/content/ViewHandler.scala
nhanitvn: Wow, great idea. It avoids join and the param numPartitions in generating randomRDD and thus has high probability to be more efficient than my approach. I think that would just erase my work, lol.

core/src/main/java/io/ddf/content/IHandleViews.java
nhanitvn: Resolved, I changed to API names :D

nhanitvn:
The change on API names leads to a lot of changes in ddf-jdbc and PE. So I will close this ticket and open another one focusing on just the logic, not name changes.

nhanitvn:
Here it is https://github.com/ddf-project/DDF/pull/328

PR #331, opened 2016-05-10
getColumn should throw exception if column does not exist

Huandao0812:

Reviewers: @phvu 

### How to test

### PR Progress
- [x] Code review is done.


phvu:
looks good.

phvu:
but lots of UT fails...

core/src/main/java/io/ddf/content/SchemaHandler.java
```diff
@@ -76,12 +76,12 @@ public int getNumColumns() {
   }
 
   @Override
-  public int getColumnIndex(String columnName) {
+  public int getColumnIndex(String columnName) throws DDFException {
     return mSchema != null ? mSchema.getColumnIndex(columnName) : -1;
```
hai-adatao: Should we change this implementation to something like:
```
if mSchema is null
    throw exception

return mSchema.getColumnIndex
```
?

core/src/main/java/io/ddf/content/SchemaHandler.java
```diff
@@ -76,12 +76,12 @@ public int getNumColumns() {
...
   }
 
   @Override
-  public String getColumnName(int columnIndex) {
+  public String getColumnName(int columnIndex) throws DDFException {
     return mSchema != null ? mSchema.getColumnName(columnIndex) : null;
```
hai-adatao: here too

core/src/main/java/io/ddf/content/ViewHandler.java
```diff
@@ -83,15 +83,13 @@ public DDF project(List<String> columnNames) throws DDFException {
     DDF projectedDDF = sql2ddf(String.format("SELECT %s FROM %%s", selectedColumns),
         String.format("Unable to project column(s) %s from table %%s", selectedColumns));
     //reserve factor information
-    List<Schema.Column> columns = this.getDDF().getSchema().getColumns();
```
hai-adatao: What is the change for this?

core/src/main/java/io/ddf/content/SchemaHandler.java
nhanitvn: Actually, -1 is never returned because the schema's getColumnIndex throws Exception for non-existing columns already :D

core/src/main/java/io/ddf/content/ViewHandler.java
nhanitvn: Just swap the way we check for Factor columns. This saves running time a bit because we loop over the projected DDF, not the original one which has more columns.

core/src/test/java/io/basic/ddf/BasicDDFTests.java
```diff
@@ -119,4 +119,10 @@ public void testDuplicatedColumnInSetColumnNames() throws DDFException {
     List<String> columnNames = Arrays.asList("row", "price", "lotsize", "bedrooms", "bathrms", "row");
     schema.setColumnNames(columnNames);
   }
+
+  @Test(expected = DDFException.class)
+  public void testGettingInexistedColumn() throws DDFException {
```
nhanitvn: Should the name be testGettingNonExistedColumn?

core/src/main/java/io/ddf/content/ViewHandler.java
Huandao0812: not only that, projectedDDF.getColumn will throw exception if column does not exist

nhanitvn:
lgtm :)

PR #333, opened 2016-05-16
Update the way we do sampling without replacement

nhanitvn:
 due to the fact that SparkSQL does not allow rand() to accept a negative seed value in order by clause

Reviewers: @hai-adatao 

### How to test

### PR Progress
- [x] Code review is done. 

Change-Id: I34ebed02c4ff615aae99a30cd68c3e1104ee3dc0

### DIFF
--- a/spark/src/main/scala/io/ddf/spark/content/ViewHandler.scala
+++ b/spark/src/main/scala/io/ddf/spark/content/ViewHandler.scala
@@ -68,9 +68,13 @@ class ViewHandler(mDDF: DDF) extends io.ddf.content.ViewHandler(mDDF) with IHand
     }
 
     if (!withReplacement) {
-      // This subquery is a workaround for Spark's inability to order rows by rand(<negative_seed>)
-      // Example: select * from table order by rand(-123)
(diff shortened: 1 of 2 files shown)


hai-adatao:
I don't really understand the fix, why this involves a sub query? Does it affect the performance of sample?  Can we just disallow negative seed or deprecate it and workaround by shifting / getting abs value instead and create ticket for BA to migrate it?

On another thought, what does this `seed` do? Does it guarantee that we will get the same result if we use the same seed? If not can we just get its abs value instead?

nhanitvn:
You cannot run this in SparkSQL "select * from table order by rand(-123)" while these sql work
- select * from table order by rand(123)
- select * from (select *, rand(-123) as rnd from table) t order by rnd
The 2nd query does not affect the performance because Spark will generate just the same number of stages like the one with no subquery.

hai-adatao:
We're fixing that by introducing a subquery, it will affect performance right? Also, my opinion is negative seed doesn't make sense, so we can just:
- use abs(seed) before using it now, and create a ticket for BA to migrate negative seed
- once BA is done, add error message to verify negative seed
- or we can just disallow negative seed in DDF now, and do abs() stuff in client

If negative seed really makes sense, then I think do an if check and only on sub query when seed is negative is better

nhanitvn:
I did a quick benchmark and the subquery does increase the running time. So I refactor to handle the negative seed case separately. Performing an abs() is not a good one, I think, though it is not that critical.

PR #340, opened 2016-05-19
Add support to scale just a subset of columns plus and inPlace flag

ducleminh:
Add support to scale a subset of columns (to both scaleMinMax and scaleStandard)
Add inPlace flag (to both scaling methods)
- https://adatao.atlassian.net/browse/PE-1774
- https://adatao.atlassian.net/browse/PE-1898
- https://adatao.atlassian.net/browse/PE-1897

Reviewers: @hai-adatao @Huandao0812 @nhanitvn @phvu @lebinh @ubolonton @zkidkid 

None

### How to test
Unit tests

### PR Progress
- [ ] Code review is done.

### DIFF
--- a/core/src/main/java/io/ddf/etl/TransformationHandler.java
+++ b/core/src/main/java/io/ddf/etl/TransformationHandler.java
@@ -30,7 +30,7 @@ public class TransformationHandler extends ADDFFunctionalGroupHandler implements
 
   @Override
   public DDF transformScaleMinMax() throws DDFException {
-    return this.transformScaleMinMax(new ArrayList<>(), Boolean.FALSE);
+    return this.transformScaleMinMax(null, Boolean.FALSE);
   }
 
   @Override
@@ -74,7 +74,7 @@ public class TransformationHandler extends ADDFFunctionalGroupHandler implements
 
   @Override
   public DDF transformScaleStandard() throws DDFException {
-    return this.transformScaleStandard(new ArrayList<>(), Boolean.FALSE);
+    return this.transformScaleStandard(null, Boolean.FALSE);
   }
 
   @Override
@@ -502,6 +502,6 @@ public class TransformationHandler extends ADDFFunctionalGroupHandler implements
   private boolean isColumnEligibleToScale(Column column, List<String> scaleRequestedColumnNames) {
     return (column.isNumeric()
             && column.getColumnClass() != ColumnClass.FACTOR
-            && ((scaleRequestedColumnNames == null) || scaleRequestedColumnNames.isEmpty() || scaleRequestedColumnNames.contains(column.getName())));
+            && ((scaleRequestedColumnNames == null) || scaleRequestedColumnNames.contains(column.getName())));
   }
 }
--- a/core/src/main/java/io/ddf/etl/TransformationHandler.java
+++ b/core/src/main/java/io/ddf/etl/TransformationHandler.java
@@ -13,6 +13,7 @@ import io.ddf.datasource.SQLDataSourceDescriptor;
 import io.ddf.exception.DDFException;
 import io.ddf.misc.ADDFFunctionalGroupHandler;
 import io.ddf.util.Utils;
+import org.apache.commons.lang.StringEscapeUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -42,29 +43,30 @@ public class TransformationHandler extends ADDFFunctionalGroupHandler implements
 
     for (int colIndex = 0; colIndex < schema.getNumColumns(); colIndex++) {
       Column col = schema.getColumn(colIndex);
+      String escapedColName = StringEscapeUtils.escapeJava(col.getName());
 
       if (!isColumnEligibleToScale(col, columns)) {
-        sqlCmdBuffer.append(col.getName()).append(",");
+        sqlCmdBuffer.append(escapedColName).append(",");
       } else {
         if (summaryArr[colIndex] == null) {
           LOG.warn("Missing column summary. transformScaleMinMax ignored for column: " + col.getName());
 
-          sqlCmdBuffer.append(col.getName()).append(" ");
+          sqlCmdBuffer.append(escapedColName).append(" ");
         } else if (Double.compare(summaryArr[colIndex].max(), summaryArr[colIndex].min()) == 0) {
           LOG.warn("max equals min. transformScaleMinMax ignored for column: " + col.getName());
 
-          sqlCmdBuffer.append(col.getName()).append(" ");
+          sqlCmdBuffer.append(escapedColName).append(" ");
         } else {
           // subtract min, divide by (max - min)
-          sqlCmdBuffer.append(String.format("((%s - %s) / %s) as %s ", col.getName(), summaryArr[colIndex].min(),
-                  (summaryArr[colIndex].max() - summaryArr[colIndex].min()), col.getName()));
+          sqlCmdBuffer.append(String.format("((%s - %s) / %s) as %s ", escapedColName, summaryArr[colIndex].min(),
+                  (summaryArr[colIndex].max() - summaryArr[colIndex].min()), escapedColName));
         }
 
         sqlCmdBuffer.append(",");
       }
     }
     sqlCmdBuffer.setLength(sqlCmdBuffer.length() - 1);
-    sqlCmdBuffer.append(" FROM ").append(schema.getTableName());
+    sqlCmdBuffer.append(" FROM ").append(StringEscapeUtils.escapeJava(schema.getTableName()));
 
     DDF newddf = this.getManager().sql2ddf(sqlCmdBuffer.toString(), this.getEngine());
     newddf.getMetaDataHandler().copyFactor(this.getDDF());
@@ -86,30 +88,31 @@ public class TransformationHandler extends ADDFFunctionalGroupHandler implements
 
     for (int colIndex = 0; colIndex < schema.getNumColumns(); colIndex++) {
       Column col = schema.getColumn(colIndex);
+      String escapedColName = StringEscapeUtils.escapeJava(col.getName());
 
       if (!isColumnEligibleToScale(col, columns)) {
-        sqlCmdBuffer.append(col.getName()).append(",");
+        sqlCmdBuffer.append(escapedColName).append(",");
       } else {
         if (summaryArr[colIndex] == null) {
           LOG.warn("Missing column summary. transformScaleStandard ignored for column: " + col.getName());
 
-          sqlCmdBuffer.append(col.getName()).append(" ");
+          sqlCmdBuffer.append(escapedColName).append(" ");
         } else if (summaryArr[colIndex].stdev() == 0) {
           LOG.warn("standard deviation equals zero. transformScaleStandard ignored for column: " + col.getName());
 
-          sqlCmdBuffer.append(col.getName()).append(" ");
+          sqlCmdBuffer.append(escapedColName).append(" ");
         } else {
           // subtract mean, divide by stdev
           sqlCmdBuffer.append(String
-                  .format("((%s - %s) / %s) as %s ", col.getName(), summaryArr[colIndex].mean(), summaryArr[colIndex].stdev(),
-                          col.getName()));
+                  .format("((%s - %s) / %s) as %s ", escapedColName, summaryArr[colIndex].mean(), summaryArr[colIndex].stdev(),
+                          escapedColName));
         }
 
         sqlCmdBuffer.append(",");
       }
     }
     sqlCmdBuffer.setLength(sqlCmdBuffer.length() - 1);
-    sqlCmdBuffer.append(" FROM ").append(schema.getTableName());
+    sqlCmdBuffer.append(" FROM ").append(StringEscapeUtils.escapeJava(schema.getTableName()));
 
     DDF newddf = this.getManager().sql2ddf(sqlCmdBuffer.toString(), this.getEngine());
     newddf.getMetaDataHandler().copyFactor(this.getDDF());
--- a/core/src/main/java/io/ddf/etl/TransformationHandler.java
+++ b/core/src/main/java/io/ddf/etl/TransformationHandler.java
@@ -5,6 +5,7 @@ package io.ddf.etl;
 import com.google.common.collect.Lists;
 import io.ddf.DDF;
 import io.ddf.analytics.Summary;
+import io.ddf.content.Schema;
 import io.ddf.content.Schema.Column;
 import io.ddf.content.Schema.ColumnClass;
 import io.ddf.content.Schema.ColumnType;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -12,70 +12,108 @@
...
+    StringBuilder sqlCmdBuffer = new StringBuilder("SELECT ");
+
+    for (int colIndex = 0; colIndex < this.getDDF().getSchema().getNumColumns(); colIndex++) {
+      Column col = this.getDDF().getSchema().getColumn(colIndex);
 
-    for (int i = 0; i < columns.size(); i++) {
-      Column col = columns.get(i);
-      if (!col.isNumeric() || col.getColumnClass() == ColumnClass.FACTOR) {
-        sqlCmdBuffer.append(col.getName()).append(" ");
+      if (!col.isNumeric() || col.getColumnClass() == ColumnClass.FACTOR
```
hai-adatao: Let's make this check a private method, dunno what we're doing here..

ddf-test/src/main/scala/io/ddf/test/it/etl/TransformationHandlerBaseSuite.scala
```diff
@@ -55,22 +55,251 @@ trait TransformationHandlerBaseSuite extends BaseSuite with Matchers {
...
+    scaledDDF.getNumColumns should be(11)
+
+    val scaledDDFSummary: Array[Summary] = scaledDDF.getSummary
+    scaledDDFSummary(0).min should be(0)
+    scaledDDFSummary(0).max should be(1)
+    scaledDDFSummary(1).min should be(1)
+    scaledDDFSummary(1).max should be(11)
+    scaledDDFSummary(4).min should be(617)
+    scaledDDFSummary(4).max should be(2107)
+    // The following column has min == max, hence should not be scaled
```
hai-adatao: Is this new logic or it was there in the beginning?

hai-adatao:
Will merge in 1.4.17, we need all PRs in downstream ready

core/src/main/java/io/ddf/etl/TransformationHandler.java
ducleminh: refactored. Here we're checking each column in DDF to see if it's eligible to be scaled

ddf-test/src/main/scala/io/ddf/test/it/etl/TransformationHandlerBaseSuite.scala
ducleminh: This is new logic added to ignore scaling if min==max as this would cause a divide-by-zero exception. At the client side, the documentation also mentions this but we don't have an explicit check in PE. (Same idea for scale standard if stdev == 0)

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -12,70 +12,106 @@
...
     super(theDDF);
     // TODO Auto-generated constructor stub
   }
 
-  @Override public DDF transformScaleMinMax() throws DDFException {
-    Summary[] summaryArr = this.getDDF().getSummary();
-    List<Column> columns = this.getDDF().getSchema().getColumns();
+  @Override
+  public DDF transformScaleMinMax() throws DDFException {
+    return this.transformScaleMinMax(new ArrayList<>(), Boolean.FALSE);
```
phvu: What's wrong with `false` instead of `Boolean.FALSE`?

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -12,70 +12,106 @@
...
-  @Override public DDF transformScaleStandard() throws DDFException {
-    Summary[] summaryArr = this.getDDF().getSummary();
-    List<Column> columns = this.getDDF().getSchema().getColumns();
+  @Override
+  public DDF transformScaleStandard() throws DDFException {
+    return this.transformScaleStandard(new ArrayList<>(), Boolean.FALSE);
+  }
 
+  @Override
+  public DDF transformScaleStandard(List<String> columns, Boolean inPlace) throws DDFException {
```
phvu: Code looks pretty similar to transformScaleMinmax. Can we refactor this somehow?

core/src/main/java/io/ddf/etl/TransformationHandler.java
ducleminh: To prevent autoboxing then autounboxing as the method expects an object Boolean

core/src/main/java/io/ddf/etl/TransformationHandler.java
ducleminh: yes, I have a ticket to add custom range scaling, will do refactoring then. 
https://adatao.atlassian.net/browse/PE-1872

phvu: don't worry. You can do it if it doesn't take much time, otherwise this is good to merge.

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -459,4 +495,10 @@ public DDF sort(List<String> columns, List<Boolean> ascending, Boolean inPlace)
       return castType(column, newType);
     }
   }
+
+  private boolean isColumnEligibleToScale(Column column, List<String> scaleRequestedColumnNames) {
+    return (column.isNumeric()
+            && column.getColumnClass() != ColumnClass.FACTOR
+            && ((scaleRequestedColumnNames == null) || scaleRequestedColumnNames.isEmpty() || scaleRequestedColumnNames.contains(column.getName())));
```
ubolonton: Why do we want to treat the empty case as "requesting all columns"? That's confusing and error-prone.

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -12,70 +12,106 @@
...
+    Summary[] summaryArr = this.getDDF().getSummary();
     // Compose a transformation query
-    StringBuffer sqlCmdBuffer = new StringBuffer("SELECT ");
+    StringBuilder sqlCmdBuffer = new StringBuilder("SELECT ");
 
-    for (int i = 0; i < columns.size(); i++) {
-      Column col = columns.get(i);
-      if (!col.isNumeric() || col.getColumnClass() == ColumnClass.FACTOR) {
-        sqlCmdBuffer.append(col.getName()).append(" ");
+    for (int colIndex = 0; colIndex < this.getDDF().getSchema().getNumColumns(); colIndex++) {
```
ubolonton: - Why don't we assign `schema` to a variable instead of calling `getSchema()` repeatedly?
- Why don't we use `schema.getColumns()`?

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -12,70 +12,106 @@
...
+      Column col = this.getDDF().getSchema().getColumn(colIndex);
+
+      if (!isColumnEligibleToScale(col, columns)) {
+        sqlCmdBuffer.append(col.getName()).append(",");
```
ubolonton: Shouldn't we handle column name escaping here, and below?

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -12,70 +12,106 @@
...
+
+          sqlCmdBuffer.append(col.getName()).append(" ");
+        } else if (Double.compare(summaryArr[colIndex].max(), summaryArr[colIndex].min()) == 0) {
+          LOG.warn("max equals min. transformScaleMinMax ignored for column: " + col.getName());
+
+          sqlCmdBuffer.append(col.getName()).append(" ");
+        } else {
+          // subtract min, divide by (max - min)
+          sqlCmdBuffer.append(String.format("((%s - %s) / %s) as %s ", col.getName(), summaryArr[colIndex].min(),
+                  (summaryArr[colIndex].max() - summaryArr[colIndex].min()), col.getName()));
```
ubolonton: Can we assign repeated expressions to variables?

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -12,70 +12,106 @@
...
+        }
+
+        sqlCmdBuffer.append(",");
       }
-      sqlCmdBuffer.append(",");
     }
     sqlCmdBuffer.setLength(sqlCmdBuffer.length() - 1);
-    sqlCmdBuffer.append("FROM ").append(this.getDDF().getTableName());
+    sqlCmdBuffer.append(" FROM ").append(this.getDDF().getTableName());
```
ubolonton: Table name should likely be escaped.

core/src/main/java/io/ddf/etl/TransformationHandler.java
ubolonton: Why does the method expect `Boolean` instead of `boolean`?

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -12,70 +12,106 @@
...
+      Column col = this.getDDF().getSchema().getColumn(colIndex);
+
+      if (!isColumnEligibleToScale(col, columns)) {
+        sqlCmdBuffer.append(col.getName()).append(",");
       } else {
-        // subtract min, divide by (max - min)
-        sqlCmdBuffer.append(String.format("((%s - %s) / %s) as %s ", col.getName(), summaryArr[i].min(),
-            (summaryArr[i].max() - summaryArr[i].min()), col.getName()));
+        if (summaryArr[colIndex] == null) {
+          LOG.warn("Missing column summary. transformScaleMinMax ignored for column: " + col.getName());
```
ubolonton: Should this be `LOG.error`?

Huandao0812:
@ducleminh: can you research to see if we could do this with Spark DataFrame, it's not good practice to generate the SQL string then pass it into SparkSQL

core/src/main/java/io/ddf/etl/TransformationHandler.java
ducleminh: No strong preference here, I followed the convention of other methods with inPlace flag.

ubolonton:
Seconded @Huandao0812's suggestion.

core/src/main/java/io/ddf/etl/TransformationHandler.java
ducleminh: - Refactored to add local schema variable.
- As schema.getColumns() clones a new list of all columns which is expensive.

core/src/main/java/io/ddf/etl/TransformationHandler.java
ducleminh: Agree, will refactor this.

ducleminh: This has been refactored in DDF, pAnalytics, PyClient, and RClient. @ubolonton @hai-adatao

core/src/main/java/io/ddf/etl/TransformationHandler.java
ducleminh: refactored.

core/src/main/java/io/ddf/etl/TransformationHandler.java
ducleminh: refactored.

core/src/main/java/io/ddf/etl/TransformationHandler.java
ducleminh: refactored.

core/src/main/java/io/ddf/etl/TransformationHandler.java
ducleminh: I'd argue using warn as here it's a check for safety, not really the right place to throw/notify this particular error.

ducleminh:
Yes, ticket raised for this: https://adatao.atlassian.net/browse/PE-2082

ddf-test/src/main/scala/io/ddf/test/it/etl/TransformationHandlerBaseSuite.scala
```diff
@@ -55,22 +55,295 @@ trait TransformationHandlerBaseSuite extends BaseSuite with Matchers {
...
+      "CRSDepTime", "ArrTime", "CRSArrTime", "UniqueCarrier", "FlightNum", "TailNum")
+    val originalDDFSummary: Array[Summary] = originalDDF.getSummary
+
+    val scaledDDF: DDF = originalDDF.Transform.transformScaleStandard(null, true)
+    scaledDDF should not be null
+    scaledDDF.getNumRows should be(31)
+    scaledDDF.getNumColumns should be(11)
+
+    val scaledDDFSummary: Array[Summary] = scaledDDF.getSummary
+    BigDecimal(scaledDDFSummary(0).mean).setScale(2, BigDecimal.RoundingMode.HALF_UP).toDouble should be(0)
```
nhanitvn: Interesting, why do we have to use BigDecimal for this check? @ducleminh

nhanitvn:
lgtm. I am going to merge this.

ddf-test/src/main/scala/io/ddf/test/it/etl/TransformationHandlerBaseSuite.scala
ducleminh: Just for rounding the double.

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -5,77 +5,119 @@
...
 
-    for (int i = 0; i < columns.size(); i++) {
-      Column col = columns.get(i);
-      if (!col.isNumeric() || col.getColumnClass() == ColumnClass.FACTOR) {
-        sqlCmdBuffer.append(col.getName()).append(" ");
+    for (int colIndex = 0; colIndex < schema.getNumColumns(); colIndex++) {
+      Column col = schema.getColumn(colIndex);
+      String escapedColName = StringEscapeUtils.escapeJava(col.getName());
+
+      if (!isColumnEligibleToScale(col, columns)) {
```
nhanitvn: @ducleminh Should we check for non-existent requested columns so that we don't have to run and return the new DDF with the same content in case of inPlace = false?

nhanitvn: No worry, it is not urgent, though.

ducleminh: @nhanitvn agree, I've added a note to this ticket to add this extra check: https://adatao.atlassian.net/browse/PE-1872

core/src/main/java/io/ddf/etl/TransformationHandler.java
```diff
@@ -5,77 +5,119 @@
...
-    StringBuffer sqlCmdBuffer = new StringBuffer("SELECT ");
+    StringBuilder sqlCmdBuffer = new StringBuilder("SELECT ");
 
-    for (int i = 0; i < columns.size(); i++) {
-      Column col = columns.get(i);
-      if (!col.isNumeric() || col.getColumnClass() == ColumnClass.FACTOR) {
-        sqlCmdBuffer.append(col.getName()).append(" ");
+    for (int colIndex = 0; colIndex < schema.getNumColumns(); colIndex++) {
+      Column col = schema.getColumn(colIndex);
+      String escapedColName = StringEscapeUtils.escapeJava(col.getName());
```
ubolonton: Shouldn't the escape rules be specific to the underlying engine, or SQL in general, not Java?

ducleminh: Good point. I'll do some more investigation if we require any escape rules for SQL queries in this ticket: https://adatao.atlassian.net/browse/PE-1872 as they're used in quite a lot of places. @ubolonton

PR #346, opened 2016-05-25
Add week UDF which performs the same task as weekofyear

nhanitvn:
- Also add weekofmonth UDF and refactor test cases

Reviewers: @hai-adatao 

### How to test

### PR Progress
- [x] Code review is done.


spark/src/main/scala/io/ddf/spark/etl/DateUDFs.scala
```diff
@@ -112,6 +112,17 @@ object DateUDF {
     }
   }
 
+  val parseWeekOfMonth: Object => Integer = {
+    (obj: Object) => {
+      val dateTime = Utils.toDateTimeObject(obj)
+      if(dateTime != null) {
+        (dateTime.getDayOfMonth() / 7) + 1
```
ubolonton: How useful is this compared to the `weekofyear` approach (week belongs to a month if more of it falls in that month)?

nhanitvn: Good point, this way of computing week of month may not be consistent with the one of weekofyear from the libraries (joda here). A week may span over two months. Also, similar to weekyear, we may need weekmonth too. Currently, we rely on joda to compute those values and joda does not offer weekofmonth. So I think we should not offer this weekofmonth UDF until there are requests from real customers.

PR #350, opened 2016-05-30
binning API now can return a factor column or an integer one

nhanitvn:
- That is done by adding **toLabels** params to allow the output column to be an integer one which contains zero-based indices instead of a factor column of range strings
- **Refactor the code, move the implementation to ABinningHandler to avoid duplicated code in spark/jdbc BinningHandler**
- Add convenient methods for binning to DDF while keeping the old one for backward compatible
- Add precision param for number of decimal digits to display in range strings (default is 2)
- Automatically search for appropriate precision value (20 tries from the specified precision value, default is 2) for nice labels
- Add validation for duplicated/non-monotonically-increasing breaks
- Binning by equal intervals and by equal frequency always use includeLowest = True despite any supplied value
- In case of binning by equal intervals, if min = max, the lower and upper bounds are extended by .1% as in Pandas and R to ensure that we can still map values to a range/value.
- Change EQUAlFREQ (lower-cased L) to EQUALFREQ (upper-cased L)

Reviewers: @hai-adatao @phvu @Huandao0812 

### How to test

### PR Progress
- [ ] Code review is done.


core/src/main/java/io/ddf/analytics/ABinningHandler.java
```diff
@@ -22,21 +30,192 @@ public ABinningHandler(DDF theDDF) {
...
+    BinningType binType = BinningType.get(binningType);
+
+    double[] theBreaks = breaks;
+
+    switch(binType) {
+      case CUSTOM:
+        if (breaks == null) throw new DDFException("Please enter valid break points");
+        // check for monotonic
+        Arrays.sort(theBreaks);
+        if (!Arrays.equals(theBreaks, breaks)) throw new DDFException("Please enter increasing breaks");
```
phvu: Does this work? `Arrays.sort()` is in-place, and `theBreaks` is just a reference to `breaks`, so after this `Arrays.equals(theBreaks, breaks)` will be true anyway?
You might need to clone `breaks` somehow.

nhanitvn: Good point, let me write a test for that.

core/src/main/java/io/ddf/analytics/ABinningHandler.java
```diff
@@ -22,21 +30,192 @@ public ABinningHandler(DDF theDDF) {
...
+        if (numBins < 2) throw new DDFException("Number of bins cannot be smaller than 2");
+        theBreaks = getQuantilesFromNumBins(colMeta.getName(), numBins);
+        includeLowest = true;
+        break;
+      case EQUALINTERVAL:
+        if (numBins < 2) throw new DDFException("Number of bins cannot be smaller than 2");
+        theBreaks = getIntervalsFromNumBins(colMeta.getName(), numBins, right);
+        includeLowest = true;
+        break;
+      default:throw new DDFException(String.format("Binning type %s is not supported", binningType));
```
phvu: nit:  new line after `default:`

core/src/main/java/io/ddf/analytics/ABinningHandler.java
```diff
@@ -22,21 +30,192 @@ public ABinningHandler(DDF theDDF) {
...
+    }
+
+    // Check for uniqueness
+    List<Double> listBreaks = Arrays.asList(ArrayUtils.toObject(theBreaks));
+    if (new HashSet<>(listBreaks).size() < theBreaks.length) {
```
phvu: One java8 way to do this is: `if (!Arrays.stream(theBreaks).allMatch(new HashSet<>()::add)) {`

core/src/main/java/io/ddf/analytics/ABinningHandler.java
```diff
@@ -22,21 +30,192 @@ public ABinningHandler(DDF theDDF) {
...
 
-    newddf.getMetaDataHandler().copyFactor(this.getDDF());
-    return newddf;
+    newDDF.getMetaDataHandler().copyFactor(this.getDDF());
+    return newDDF;
+  }
+
+  private double[] getIntervalsFromNumBins(String colName, int numBins, Boolean right) throws DDFException {
+    DDF ddf = this.getDDF();
+    double[] minMax = new double[] {ddf.getVectorMin(colName), ddf.getVectorMax(colName)};
```
phvu: nit: unnecessary array

hai-adatao:
@nhanitvn can you help put all these tests in `ddf-test/BinningHandlerBaseSuite`, and then implement spark test suite instead? Check out `JoinHandlerBaseSuite` and `JoinHandlerSuite` for an example.
Also, would appreciate if you take a look at the tests in `ddf-test/BinningHandlerBaseSuite`, they're all ignored, if they're not valid anymore please help remove it, or refactor it otherwise.

hai-adatao:
Let's try to get this in 1.4.17

core/src/main/java/io/ddf/analytics/ABinningHandler.java
nhanitvn: Wow, concise code. However it seems to be magical and difficult to figure out how it achieve the goal. May be we just keep the current one.

PR #351, opened 2016-05-31
Feature/gc ddf

Huandao0812:
DDFCache is backed by a guava's Loading cache, 
https://github.com/google/guava/wiki/CachesExplained
The eviction policy is LRU
For DDF cache we use both size-based eviction and time eviction:
  + cache's capacity is set by MAX_NUMBER_OF_DDFS_IN_CACHE, the cache will evict DDF that haven't been used recently.
  + Evict DDF after DDF_TIME_TO_IDLE_SECONDS seconds have passed since it last accessed.
  + Restore DDF by using DDFManager.restore function if there's request for DDF after it was evicted

Reviewers: @zkidkid @nhanitvn @phvu @lebinh @ubolonton 

### PR Progress
- [x] Code review is done.


core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,90 @@
...
-  public boolean hasDDF(UUID uuid) {
-    DDF ddf = mDDFs.get(uuid);
-    return ddf != null;
+  public DDF getDDF(UUID uuid) {
+    return mDDFCache.getUnchecked(uuid);
   }
 
   public DDF getDDFByName(String name) throws DDFException {
-    for(DDF ddf: mDDFs.values()) {
+    for(DDF ddf: this.listDDFs()) {
```
keimoon: Should we have another ConcurrentHashMap here to speed up this method

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,90 @@
...
+      }
+    });
+  }
+
+  private boolean containsDDF(UUID uuid) {
+    return mDDFCache.asMap().containsKey(uuid);
+  }
 
-  private Map<UUID, DDF> mDDFs = new ConcurrentHashMap<UUID, DDF>();
   private Map<String, UUID> mUris = new ConcurrentHashMap<String, UUID>();
```
phvu: This is not used to look up anywhere. Maybe just get  rid of it?

core/src/main/java/io/ddf/DDFCache.java
phvu: 4000 items is not that big, I think we can go on with this until we see performance degrades.
Using another hashmap will add the burden of maintaining consistency between the hashmap and mDDFCache, which may be tricky because this class will be used in different threads.

core/src/main/java/io/ddf/DDFCache.java
phvu: Oh no scratch that. Used in getDDFByUri.

phvu:
lgtm

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,90 @@
...
+        } catch (Exception e) {
+          mLog.error(String.format("Error restoring DDF %s, error = %s", uuid, ExceptionUtils.getStackTrace(e)));
+          throw new DDFException(String.format("DDF with uuid %s does not exist", uuid));
+        }
+      }
+    });
+  }
+
+  private boolean containsDDF(UUID uuid) {
+    return mDDFCache.asMap().containsKey(uuid);
```
lebinh: this is potentially costly, depends on what `asMap` is doing. Maybe `getIfPresent` and check for not null result?

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,90 @@
...
+
+  private DDFManager mDDFManager;
+
+  public DDFCache(DDFManager manager) {
+    mDDFManager = manager;
+    /**
+     * Maximum number of DDFs allowed in cache, DDFs will be evicted by an approximate LRU algorithm
+     */
+    Long maxNumberOfDDFs = Long.valueOf(Config.getGlobalValue(Config.ConfigConstant.MAX_NUMBER_OF_DDFS_IN_CACHE));
+
```
ducleminh: One thought, when we cache a DDF, is the underlying data also cached? if it is, as the size of DDF can be very large, we could hit OOM even before the number of DDF limit in the cache is hit.

nhanitvn:
lgtm

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -68,19 +106,18 @@ public synchronized void setDDFName(DDF ddf, String name) throws DDFException {
...
       ddf.setUUID(uuid);
-      mDDFs.put(uuid, ddf);
-      if(ddf.getUri() != null) {
-        mUris.remove(ddf.getUri());
-        mUris.put(ddf.getUri(), ddf.getUUID());
+      if(prevUUID != null) {
+        mDDFCache.invalidate(prevUUID);
+        mDDFCache.put(uuid, ddf);
+        if(ddf.getUri() != null) {
+          mUris.remove(ddf.getUri());
```
lebinh: this isn't necessary as `put` will replace/remove the current value anyway

core/src/main/java/io/ddf/DDFCache.java
ubolonton: `asMap` seems to be used to avoid changing the cache's state. This seems to be used only when restoring a DDF, so that wouldn't actually be a concern. On the other hand, that also means that it's still ok if it's not too efficient.

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,90 @@
...
-    DDF ddf = mDDFs.get(uuid);
-    return ddf != null;
+  public DDF getDDF(UUID uuid) {
+    return mDDFCache.getUnchecked(uuid);
   }
 
   public DDF getDDFByName(String name) throws DDFException {
-    for(DDF ddf: mDDFs.values()) {
+    for(DDF ddf: this.listDDFs()) {
       if(!Strings.isNullOrEmpty(ddf.getName()) && ddf.getName().equals(name)) {
```
ducleminh: isNullOrEmpty is redundant here, it's done by setDDFName.

ubolonton:
- This creates the issue of important not-used-recently DDFs being prioritized for removal over recent temporary DDFs. While it is probably ok for the use case we are fixing, it will likely cause spikes of performance degradation for other use cases.
- Can we make this a handler so that PE can customize it? For example it can use a different caching strategy that takes a "cache"/"pin" flag into account, which would be specific to PE's use cases, not DDF core.

core/src/main/java/io/ddf/DDFCache.java
ubolonton: Caching of underlying data is managed by the engine (Spark, which already does disk-spill). We don't have to worry about that here.

Huandao0812: @ducleminh this is more for managing DDF in PE's memory, Spark has its own memory management

Huandao0812:
@ubolonton we could extends and override the class, one simple solution is adding a cache flag to signal don't remove this DDF

core/src/main/java/io/ddf/DDFCache.java
Huandao0812: put(uuid, ddf) will remove value that associated key if any, it doesn't remove value associated why prevUUID

lebinh: this is `remove` and `put` for the same URI

ducleminh:
@ubolonton @Huandao0812 or don't cache the temporary DDFs if we can work out whether they are temporary.

nhanitvn:
retest this please

hai-adatao:
Hold on, don't merge this one yet

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,90 @@
...
+
+    /**
+     * The number of seconds that the DDF will live in cache since it last used(meaning inserted or accessed)
+     */
+    Long timeToIdle = Long.valueOf(Config.getGlobalValue(Config.ConfigConstant.DDF_TIME_TO_IDLE_SECONDS));
+
+    mLog.info(String.format("Maximum number of ddfs in cache %s", maxNumberOfDDFs));
+    mLog.info(String.format("DDF's time to idle = %s seconds", timeToIdle));
+    mDDFCache = CacheBuilder.newBuilder().
+        maximumSize(maxNumberOfDDFs).expireAfterAccess(timeToIdle, TimeUnit.SECONDS).
```
zkidkid: @Huandao0812: MaximumSize usually use for size-base eviction, expireAfterAccess other hand use with time-base eviction, I don't think it's good to use both. 
Only maximumSize is enough, it will use LRU to evict by default.

core/src/main/java/io/ddf/DDFCache.java
zkidkid: @lebinh : I think asMap is fine & better than getIfPresent.

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,90 @@
...
+        recordStats().removalListener(new DDFRemovalListener())
+        .build(new CacheLoader<UUID, DDF>() {
+      @Override public DDF load(UUID uuid) throws Exception {
+        try {
+          mLog.info(String.format("restoring ddf %s", uuid));
+          return mDDFManager.restoreDDF(uuid);
```
zkidkid: mDDFManager.restoreDDF(uuid) currently not implement yet, thus It will lead to not consistent in DDFManager behavior.

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,90 @@
...
-    mDDFs.remove(ddf.getUUID());
+    mDDFCache.invalidate(ddf.getUUID());
     if(ddf.getUri() != null) {
       mUris.remove(ddf.getUri());
     }
   }
 
   public DDF[] listDDFs() {
-    return this.mDDFs.values().toArray(new DDF[] {});
+    return mDDFCache.asMap().values().toArray(new DDF[]{});
```
zkidkid: I suggest to change this signature to Iterator.
Currently we use listDDFs() in a loop condition:
for(ddf : listDDFS()) { ...} so it's better to return Iterator here.

core/src/main/java/io/ddf/DDFCache.java
Huandao0812: can you elaborate why you don't think it's good to use both?

core/src/main/java/io/ddf/DDFCache.java
zkidkid: @Huandao0812 UUID is right to remove prevUUID then put new.
but Uri is not as Binh said.
```
_mUris.remove(ddf.getUri());_
mUris.put(ddf.getUri(), ddf.getUUID());
```

core/src/main/java/io/ddf/DDFManager.java
```diff
@@ -158,7 +158,11 @@ public abstract DDF transferByTable(UUID fromEngine, String tableName) throws
...
    * List of existing DDFs, only in memory one.
    */
-  protected DDFCache mDDFCache = new DDFCache();
+  protected DDFCache mDDFCache = new DDFCache(this);
+
+  public DDFCache getDDFcache() {
+    return this.mDDFCache;
+  }
 
   protected Map<String, IModel> mModels = new ConcurrentHashMap<String, IModel>();
```
zkidkid: it should be better to declare variable at top of class.

core/src/main/java/io/ddf/DDFCache.java
zkidkid: We want to keep DDF as much as possible so we will only evict when total ddf exceed our cache capacity.
for example, we set expire in 1 day, so in weekend we will evict all cached ddf whether we also have 4000 slot to keep.

Huandao0812: good point, how do others think? @phvu @ubolonton @nhanitvn @lebinh

Huandao0812:
I change the DDFCache to keep DDF with name on a separate cache, let's go with this first, a separate flag to signal long live versus ephemeral can be add later. Please review @nhanitvn @phvu @ubolonton @zkidkid @lebinh

nhanitvn:
That would be a better approach. Anyhow, we have no-named DDFs and named DDFs.

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -13,47 +17,62 @@
...
+  /**
+   * Keeping DDF without name
+   */
+  private Cache<UUID, DDF> mDDFCache;
+
+  /**
+   * Keeping DDF that has name in a separate cache, so eviction on DDF without name
+   * does not affect DDF with name
+   */
+  private Cache<UUID, DDF> mDDFwithNameCache;
```
ubolonton: Can we have a short paragraph that describes why we have 2 caches here, together with a `TODO` note asking for future improvements?

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -94,14 +170,19 @@ public DDF getDDFByName(String name) throws DDFException {
   }
 
   public synchronized void setDDFName(DDF ddf, String name) throws DDFException {
-    if(!Strings.isNullOrEmpty(name)) {
-      if(!Strings.isNullOrEmpty(ddf.getName())) {
-        this.mUris.remove(ddf.getUri());
+    synchronized (ddf) {
```
ubolonton: Why do we need both this block and the method to be `synchronized`?

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -62,26 +81,83 @@ public void addDDF(DDF ddf) throws DDFException {
...
+          } else {
+            mDDFwithNameCache.put(ddf.getUUID(), ddf);
+          }
+        } finally {
+          lock.unlock();
+        }
+      } else {
+        throw new DDFException("Timeout getting DDF");
+      }
+    } catch (InterruptedException e) {
```
ubolonton: - Shouldn't this be caught only for `lock.tryLock`?
- Does this mean that `addDDF` will successfully return when the thread is interrupted? Isn't that a bad thing?

Huandao0812: my mistake, need to throw exception here

core/src/main/java/io/ddf/DDFCache.java
ubolonton: This class has several granularities of lock. Can we have a short description of its locking strategy at the beginning?

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,166 @@
...
+
+    /**
+     * The number of seconds that the DDF will live in cache since it last used(meaning inserted or accessed)
+     */
+    Long timeToIdle = Long.valueOf(Config.getGlobalValue(Config.ConfigConstant.DDF_TIME_TO_IDLE_SECONDS));
+
+    mLog.info(String.format("Maximum number of ddfs in cache %s", maxNumberOfDDFs));
+    mLog.info(String.format("DDF's time to idle = %s seconds", timeToIdle));
+
+    mDDFCache = CacheBuilder.newBuilder().
```
zkidkid: Let move cache strategy to config instead then we could easily switch cache strategy (time-base, size-base or both) 
  `mDDFwithNameCache = CacheBuilder.from("maximumSize=200,expireAfterWrite=2m")
        .recordStats()
        .removalListener(new DDFRemovalListener())
        .build();`

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,166 @@
...
+  }
+
+  private boolean containsDDF(UUID uuid) {
+    return (mDDFCache.asMap().containsKey(uuid) || mDDFwithNameCache.asMap().containsKey(uuid));
+  }
 
-  private Map<UUID, DDF> mDDFs = new ConcurrentHashMap<UUID, DDF>();
   private Map<String, UUID> mUris = new ConcurrentHashMap<String, UUID>();
 
   public void addDDF(DDF ddf) throws DDFException {
```
zkidkid: Your idea is locking on uuid of ddf which could change wherever on user code space so it doesn't (or couldn't) lock here.

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,166 @@
...
+    }
+  }
+
+  public CacheStats getCacheStats() {
+    return mDDFCache.stats().plus(mDDFwithNameCache.stats());
   }
 
-  public void removeDDF(DDF ddf) throws DDFException {
-    mDDFs.remove(ddf.getUUID());
+  public synchronized void removeDDF(DDF ddf) throws DDFException {
```
zkidkid: no need to synchronized here.
`id = ddf.getUUID()
if(id != null){
   mDDFCache.invalidate(id)
   mDDFNameCache.invalide(id)
}
`

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,166 @@
...
+    mDDFCache.invalidate(ddf.getUUID());
+    mDDFwithNameCache.invalidate(ddf.getUUID());
     if(ddf.getUri() != null) {
       mUris.remove(ddf.getUri());
     }
   }
 
   public DDF[] listDDFs() {
```
zkidkid: listDDFs should return iterator. Incase of user want to use as List they could create it base on a interator (or readonly iterator)

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -2,52 +2,166 @@
...
-    DDF ddf = mDDFs.get(uuid);
-    return ddf != null;
+  /**
+   *  Try to get DDF from both mDDFCache and mDDFWithNameCache
+   *  if DDF is not present in both, do a restore and put it into the respective cache
+   * @param uuid
+   * @return
+   * @throws DDFException
+   */
+  public DDF getDDF(UUID uuid) throws DDFException {
```
zkidkid: Again, getDDF don't need to lock. Cache itself has threadsafe already.

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -56,31 +170,39 @@ public DDF getDDFByName(String name) throws DDFException {
   }
 
   public synchronized void setDDFName(DDF ddf, String name) throws DDFException {
```
zkidkid: I don't think it's behavior of DDFCache, I prefer to follow SRP.

core/src/main/java/io/ddf/DDFCache.java
Huandao0812: no, we have two caches here, also what if two threads trying to getDDF and restore the same DDF at the same time

spark/src/test/scala/io/ddf/spark/test/SparkDDFTests.scala
```diff
@@ -68,6 +68,11 @@ with TransformationHandlerBaseSuite with ViewsHandlerBaseSuite {
     assert(ddf1.getNumColumns == ddf2.getNumColumns)
   }
 
+  test("clean up ddf") {
```
hai-adatao: What does this test suppose to do? It asserts nothing

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -92,4 +214,38 @@ public DDF getDDFByUri(String uri) throws DDFException {
...
+            mLog.info(String.format("Removing DDF %s, name = %s", ddf.getUUID(), ddf.getName()));
+          } else {
+            mLog.info(String.format("Removing DDF %s", ddf.getUUID()));
+          }
+          ddf.cleanup();
+        }
+      }
+    }
+  }
+//
```
hai-adatao: Remove this block

core/src/main/java/io/ddf/DDFCache.java
zkidkid: Because you don't use CacheLoader, let use it & put restore function into it instead.
It guarantee that restore is threadsafe.

Huandao0812: No, I need two different caches, if I use cache loader, how does it guarantee to put the restored DDF into the correct cache

zkidkid: Make DDFManager#restore(UUID) be threadsafe.

zkidkid: Let think about this scenario:
Your thread lock on uuid, then restore
Another user execute "RestoreDDF" api -> call to SparkDDFManager#Restore
-> exist two version of ddf with same UUID.

Huandao0812: We should remove the RestoreDDF api and make DDFManager.restore private package, so it could only be called from DDFCache

Huandao0812: Make DDFManager#restore to be thread safe doesn't help either, 2 different thread calling DDFCache.getDDF could restore a DDF multiple times back to back.

zkidkid: Agree, if you couldn't use CacheLoader, you will have to provide an other load function to update your two cache. That function is threadsafe, not locking on getDDF function like that.
let say:
```
getDDF(UUID id)
 ddf = priorityCache.getIfPresent( id )
 if( ddf is null)
   ddf = normalCache.getIfPresent( id)
if (ddf is still null)
   ddf = _custom_cache_loader( id )
if(ddf is still null) throw new NotFoundException( ...)
return ddf
```

Huandao0812: two threads can still go into the line
` ddf = _custom_cache_loader(id)` 
at the same time, one will go into the function, the other will wait then go into the function after the first one is done, you will end up running the _custom_cache_loader 2 times

zkidkid: The 2nd thread will check if uuid has restored already or not, if yes then it will do nothing.

core/src/main/java/io/ddf/DDFCache.java
```diff
@@ -161,46 +158,56 @@ public DDF getDDF(UUID uuid) throws DDFException {
...
+    if (this.containsDDF(uuid)) {
       throw new DDFException(String.format("DDF with uuid %s already exists", uuid));
     } else {
       UUID prevUUID = ddf.getUUID();
       ddf.setUUID(uuid);
-      if(prevUUID != null) {
-        if(ddf.getName() == null) {
+      if (prevUUID != null) {
+        if (ddf.getName() == null) {
           mDDFCache.invalidate(prevUUID);
```
zkidkid: why do we need to invalidate prevUUID here ?
as this time, UUID isn't in your cache anymore.

PR #358, opened 2016-06-03
temporary fix for S3DDF/HDFSDDF not recognizing lower-case format names

ubolonton:
Temporary fix for S3DDF and HDFSDDF not recognizing lower-case format names.
E.g. It treated `orc` as unknown format, thus falling back to the default of `csv`.

Reviewers: @PangZhi @Huandao0812 @hai-adatao 

No

### How to test
UT
@hai-adatao We need to add more client-side tests for cases where specified format does not agree with file extensions (though we'd better remove the automatic format inference).

### PR Progress
- [X] Branch is in format `prefix/description` (see [this](http://www.guyroutledge.co.uk/blog/git-branch-naming-conventions/))
- [x] Code review is done.

### DIFF
--- a/hdfs/src/main/java/io/ddf/hdfs/HDFSDDF.java
+++ b/hdfs/src/main/java/io/ddf/hdfs/HDFSDDF.java
@@ -53,20 +53,21 @@ public class HDFSDDF extends DDF {
         }
         // Check directory or file.
         HDFSDDFManager hdfsDDFManager = this.getManager();
+        mIsDir = hdfsDDFManager.isDir(this);
         // Check dataformat.
         if (options != null && options.containsKey("format")) {
             try {
                 String format = options.get("format").toUpperCase();
                 format = format.equals("PARQUET") ? "PQT" : format;
-                mDataFormat = DataFormat.valueOf(options.get("format"));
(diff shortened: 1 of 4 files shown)


hdfs/src/main/java/io/ddf/hdfs/HDFSDDF.java
```diff
@@ -53,20 +53,21 @@ private void initialize() throws DDFException {
...
         // Check dataformat.
         if (options != null && options.containsKey("format")) {
             try {
                 String format = options.get("format").toUpperCase();
                 format = format.equals("PARQUET") ? "PQT" : format;
-                mDataFormat = DataFormat.valueOf(options.get("format"));
+                mDataFormat = DataFormat.valueOf(format);
             } catch (IllegalArgumentException e) {
-                mIsDir = hdfsDDFManager.isDir(this);
+                // TODO: Disable automatic format choosing, or put it under a convenience flag.
```
PangZhi: why we remove isDir check here @ubolonton ?

ubolonton: It's moved above so that it always run. Otherwise it's null when format is specified in `options`.

PR #365, opened 2016-06-24
[PE-2181] Support wildcard and multi-paths

PangZhi:

Able to use wildcard and list in path for ingestion

https://adatao.atlassian.net/browse/PE-2181
Reviewers: 
- Main reviewer: @hai-adatao @ubolonton @zkidkid 
### PR Progress

- [x] Main reviewer approved
- [ ] Optional reviewer approved


ubolonton:
@PangZhi @hai-adatao This needs some more improvements.
- S3 path starting with `/` does not seem to be handled correctly (empty bucket).
- There's no test for paths with wildcard.
- There's unused code.

I'll still merge it into `feature/cache-ddf` first.

spark/src/main/java/io/ddf/spark/SparkDDFManager.java
```diff
@@ -272,8 +272,23 @@ public DDF copyFrom(DDF ddf) throws DDFException {
       if (ddf instanceof S3DDF) {
         S3DDF s3DDF = (S3DDF) ddf;
         S3DataSourceCredentials cred = s3DDF.getManager().getCredential();
-        uri = String.format("s3a://%s:%s@%s/%s", cred.getAwsKeyID(), cred.getAwsScretKey(),
-            s3DDF.getBucket(), s3DDF.getKey());
+        if (s3DDF.getPath() != null) {
+          String[] fileList;
+          String path = s3DDF.getPath();
+          fileList = path.split(",");
```
ubolonton: This wouldn't work against pattern like `adatao-sample-data/test/parquet/wildcard/{part=1,part=2}/`.
We should add minimal processing to paths.
If we want to support multiple paths, make user code explicitly pass a list of paths instead.

ubolonton: Spark supports passing `paths` option, or `Seq<String>` as paths in `load`.

PangZhi: adatao-sample-data/test/parquet/wildcard/{part=1,part=2}/ will help users, but it also brings some trouble. for example, should we support adatao-sample-data/test/parquet/{date1, date2}/{part=1, part=2}?If it's in arbitary format, it's really difficult for us to handle correctly. I suggest we only allow the easiest "full path, full path"

PangZhi: @ubolonton So we should add a pyclient api using list of paths as input

ubolonton: That pattern is already supported by Spark/Hadoop. The issue is that since we additionally separate the protocol and the FS path parts of the uri in our APIs, we can only easily support either of:
- `full path, full path`: by keeping splitting by comma.
- `path/{subpath1, subpath2}/`: by not doing any additional processing on the path.

If we support passing a list of paths, the second one would be more useful. If we don't, the first one is more useful.

To support list of paths:
- PyClient/RClient accepts either a string or a list of string
- PE HTTP APIs accept either `path` or `paths` parameters
- DDF core supports `paths` when constructing DDFs, deprecating `path`

Huandao0812:
lgtm

PR #369, opened 2017-05-30
Remove MySQL, Rserve, REngine dependencies

Huandao0812:


lebinh:
good to see some activity here again 👍

